PR #13, opened 2024-02-29
[WIP] Adds code to structure DECALS images according to baseline

EiffL:
The `build_parent_sample.py` script is functional but takes a very long time. Here is for instance how I launch it on Rusty: 
```bash
$ python build_parent_sample.py /mnt/ceph/users/polymathic/external_data/astro/DECALS_Stein_et_al/ /mnt/ceph/users/polymathic/AstroPile/decals --num_processes=16
```
Just processing the north side is going to take about 20 hours...

I haven't had time yet to adjust the HF dataset script file accordingly, but that's the next step.


EiffL:
I've tested this PR and it generates the reformated files fine on Rusty. I had to do a few hacks because the southern sample is wonky, with indices skipped and out of order, or even duplicated.

It takes 27 hours to reformat north, and an estimated 70 hours for the south

PR #20, opened 2024-03-26
VIPERS download and standardization script

pmelchior:
This PR adds a download and reformat script for VIPERS DR2


scripts/vipers/build_parent_sample.py
```diff
@@ -0,0 +1,150 @@
...
+
+    # Loop through the header keys and add them to the results dictionary
+    for key in HEADER_KEYS:
+        results[key] = float(header[key])
+    
+    # Add the spectrum data to the results dictionary
+    results['spectrum_flux'] = data['FLUXES'].astype(np.float32)
+    results['spectrum_wave'] = data['WAVES'].astype(np.float32)
+    results['spectrum_ivar'] = data['NOISE'].astype(np.float32)
+    results['spectrum_mask'] = data['MASK'].astype(np.float32)
```
lhparker1, line 76: We will want to normalize these values somehow. What are the units of the fluxes/waves/noise/mask? How to move these to standardized units?

scripts/vipers/vipers.py
```diff
@@ -0,0 +1,156 @@
...
+                for k in keys:
+                    # Extract the indices of requested ids in the catalog 
+                    i = sort_index[np.searchsorted(sorted_ids, k)]
+
+                    example = {
+                        "spectrum": {
+                            "flux": data["spectrum_flux"][i],
+                            "ivar": data["spectrum_ivar"][i],
+                            "lambda": data["spectrum_wave"][i],
+                            "mask": data["spectrum_mask"][i]
```
lhparker1, line 146: If we don't normalize above, we probably want to while loading in the data

lhparker1 (changes requested):
Refactored the code to fit into the current healpix ordering + huggingface dataset format. Open questions on how to normalize the data with appropriate units are in comments in the files.

lhparker1 (approved):

scripts/vipers/vipers.py
```diff
@@ -0,0 +1,156 @@
...
+                               description="VIPERS Full Catalog")
+    ]
+
+    DEFAULT_CONFIG_NAME = "all"  # It's not mandatory to have a default configuration. Just use one if it make sense.
+
+    def _info(self):
+        """Defines the dataset info."""
+        features = datasets.Features(
+            {
+                "spectrum": Sequence({
```
EiffL, line 89: @lhparker1 it looks good, can you just use this formatting of the data to avoid Sequences:
https://github.com/AstroPile/AstroPile_prototype/blob/61e9c9a2928ebf5dd3b3778ccf50345e8b0ea852/scripts/sdss/sdss.py#L132

EiffL (changes requested):
Looking good, one small thing to change + can you add a test script?

EiffL (approved):
Excellent! Thanks a lot Liam!

PR #21, opened 2024-03-26
Add initial APOGEE DR17 support

henrysky:
This PR adds a download and reformat script for APOGEE DR17.

Compared to galaxy spectra, this dataset also added pseudo-continuum normalized spectra on top of flux spectra and spectra here are always in rest-frame.

Related to issue https://github.com/AstroPile/FlatironMeeting2024/issues/12 and https://github.com/AstroPile/FlatironMeeting2024/issues/19


scripts/apogee/apogee.py
```diff
@@ -0,0 +1,195 @@
...
+    @classmethod
+    def _info(self):
+        """Defines the features available in this dataset."""
+        # Starting with all features common to spectral dataset
+        features = {
+            "spectrum": Sequence(
+                {
+                    "flux": Value(dtype="float32"),
+                    "ivar": Value(dtype="float32"),
+                    # "lsf_sigma": Value(dtype="float32"),
```
EiffL, line 89: is it that the lsf is constant? and/or the lsf is not very important here?

For uniformity, we can still have an `lsf_sigma` entry, but set it to a constant value. The main reason for doing that is that this way, a neural network built to support lsf information will also understand this type of data.

scripts/apogee/build_parent_sample.py
```diff
@@ -0,0 +1,283 @@
...
+        return fullfilename
+    else:
+        try:
+            urllib.request.urlretrieve(urlstr, fullfilename)
+            return 0
+        except urllib.error.HTTPError as emsg:
+            return 1  # error code
+
+
+def selection_fn(base_path, catalog):
```
EiffL, line 100: can you move this selection function to the top of the file, to make it really easy for people to find?

scripts/apogee/build_parent_sample.py
```diff
@@ -0,0 +1,283 @@
...
+        raw_ivar = np.zeros(8575)
+        mask_spec = np.zeros(8575)
+
+    # Load the combined spectra file
+    if os.path.exists(continuum_filename) and continuum_filename != 1:
+        hdus = fits.open(continuum_filename)
+        continuum_flux = hdus[1].data
+        continuum_ivar = 1 / hdus[2].data ** 2
+    else:
+        continuum_flux = np.zeros(8575)
```
EiffL, line 141: does this mean that we can have empty entries? Can we find those at the catalog level and remove them with some sort of selection?

EiffL (commented):
Looking really good already! I have not yet tried to run it on my machine to see if it worked, just had a few questions first

scripts/apogee/build_parent_sample.py
henrysky: Actually the selection function have filtered out those empty entries so I have removed these codes now

scripts/apogee/build_parent_sample.py
henrysky: done

scripts/apogee/apogee.py
henrysky: I have added a rough estimate on the lsf. lsf is not important here.

scripts/apogee/apogee.py
```diff
@@ -0,0 +1,200 @@
...
+    "radial_velocity",
+    "restframe",
+]
+
+# Features that correspond to ugriz fluxes
+_FLUX_FEATURES = []
+_BOOL_FEATURES = []
+
+
+class SDSS(datasets.GeneratorBasedBuilder):
```
EiffL, line 66:
```suggestion
class APOGEE(datasets.GeneratorBasedBuilder):
```

henrysky: I have changed the class name

henrysky:
Thanks for the comments. I have pushed changes and seems to have fixed the issue. Looking at the failing CI test, it has failed at hsc dataset not APOGEE.

EiffL (approved):
This is great! Everything seems to work! Thanks @henrysky !

PR #23, opened 2024-03-26
DESI-LS: Add functions to create catalog and download images

mwalmsley:
Add two new scripts for DECaLS (aka DESI-LS as we include sister surveys):
- Query LS-DR10 TAP database to grab RA/Dec and key columns. Save as .parquet
- Given RA/Dec etc., construct a query url for the cutout server to grab the fits image.

Easy arguments for pixel size, depth and bands. TBD among community.

Note: astropile currently has DESI-LS images via Stein's HDF5 export. This is an alternative to recreate those images with e.g. different native pixel sizes, as requested by @EiffL


scripts/decals/build_galaxy10_sample.py
mhuertascompany, line 1: change the filename to "build_parent_sample.py" for consistency with other datasets

scripts/decals/build_galaxy10_sample.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def pack_to_hdf5(df: pd.DataFrame, save_loc: Path):
+    # save to hdf5
+    with h5py.File(save_loc, 'w') as f:
+ 
+        for col in ['ra', 'dec', 'label']:
+            f.create_dataset(col, data=df[col].values)
+
+        f.create_dataset('image_array', (len(df), 3, 224, 224))  # no values yet
```
mhuertascompany, line 44: the image size is hard coded - perhaps try to make it clear somewhere in the code with a global variable and replace all numerical values by the global variable name

scripts/decals/decals.py
mhuertascompany, line 1: why is the image size here 152?

scripts/decals/create_images_from_cutout_server.py
```diff
@@ -0,0 +1,40 @@
...
+    url = "http://legacysurvey.org/viewer/fits-cutout?" + urlencode(params)
+
+    return url
+
+
+if __name__ == '__main__':
+    url = get_download_url(
+        ra=150.0,
+        dec=2.0,
+        pixel_size=424,
```
mhuertascompany, line 36: another image size - please homogenize

scripts/decals/build_galaxy10_sample.py
```diff
@@ -0,0 +1,125 @@
...
+    # read directly into astropy and save as ,fits.gz
+    with BytesIO(response.content) as f:
+        astropy_fits = fits.open(f)
+        astropy_fits.writeto(save_loc, overwrite=True, output_verify='exception')
+
+def main():
+
+    logging.basicConfig(level=logging.INFO)
+
+    base_dir = Path('/Users/user/repos/zoobot-foundation/foundation/datasets/custom_downstream/decals10')
```
mhuertascompany, line 90: check this local paths. are they needed?

scripts/decals/build_galaxy10_sample.py
```diff
@@ -0,0 +1,125 @@
...
+    save_dir = base_dir / 'images'
+    debug = True
+    n_processes = 4
+    overwrite = False
+    
+    # https://huggingface.co/datasets/mwalmsley/galaxy10_decals/tree/main
+    # includes h5_index
+    train_df = pd.read_parquet(base_dir / 'decals10_train.parquet')
```
mhuertascompany, line 98: is is downloading the data?

baselines/dataset_utils.py
mhuertascompany, line 1: since the priority for now is to have the datasets ready, I'd recommend to 2 different PRs. One for the data and the other for the baselines. I have not reviewed the baseline part.

mhuertascompany (changes requested):

EiffL:
Thanks both, I'll have a go at implementing @mhuertascompany 's comments

EiffL:
Ok, I did a lot of digging, I think I know how to generate the cutouts directly from the tiles saved at NERSC, that will be much much faster than going through the cutout API, and will even avoid having to transfer all the native data over. 

I'm setting up the scripts, and we will use our polymathic hours at NERSC to execute the job. I'll open a new PR for that part of the problem. Once that works, we can remove from this PR the data download part and concentrate on the baseline code.

mwalmsley:
Hello. I will go ahead and remove the cutout download part. Please send comments on the baseline part.

mwalmsley:
Closing for a fresh baseline PR, using GZ10 
See #47

PR #18, opened 2024-03-25
initial draft of IFU datamodel

havok2063:
This PR is an initial draft for a datamodel describing inputs of IFU datasets.  Full description in DESIGN.md.  It proposes basic fields about the source observation, a `provenance` dictionary, `spaxels` as a list of individual spectra at each spatial position, and optionally `images` as a list of associated image data, e.g. reconstructed images or derived analysis maps.   

Each element in `spaxels` is an object with similar fields to the `spectrum` datamodel, plus spatial position coordinates. 
Each element in `images` is an object with a label of the type of data content with data, and optional uncertainty and mask arrays. 

```
{
  "provenance": {"project": "SDSS", "survey": "MaNGA", release: "DR17"},
  "observation_id": 111111,
  "ra": 124.,
  "dec": 0.,
  "z": 0.01,
  "z_err": 1e-5,
  "spaxel_size": 0.5",
  "spaxels": [{'flux': array(5000), 'ivar': array(5000), 'lsf_sigma': array(5000), 'lambda': array(5000), 'x': 0, 'y': 0, 'extra': {}},
  ...
  ],
  "images": [
    {"label": "reconstructed_u", "filter": "u", "data": array(34, 34), "err": null, "mask": null},
    {"label": "reconstructed_psf_u", "filter": "u", "data": array(34, 34), "err": null, "mask": null},
    ...
    {"label": "emline_gflux_6564", "data": array(34, 34), "ivar": array(34, 34), "mask": array(34, 34)},
    {"label": "stellar_vel", "data": array(34, 34), "ivar": array(34, 34), "mask": array(34, 34)},
    {"label": "d4000", "data": array(34, 34), "ivar": array(34, 34), "mask": array(34, 34)}
  ]
}
```

PR #28, opened 2024-03-27
Add spectra schema doc

maja-jablonska:


EiffL:
This is great, thanks @maja-jablonska for adding this, could you add it to the design.md document for now? when/if the document becomes too big, we can split it into a better documentation :-)

mb010:
@maja-jablonska did we stick to these schema? I.e. should I be merging them, or was it preliminary and likely doesn't apply anymore?

mb010:
#18 did an initial pass at this, and that seems to be the documentation that has been adopted by the rest of the modality teams.

PR #29, opened 2024-03-27
Add Gaia DR3

al-jshen:
Adds a HF dataset for the subset of Gaia DR3 for which mean BP/RP spectra are available (~220 million stars). Includes astrometry, photometry, BP/RP spectra(l coefficients), photometrically-derived stellar parameters, and correction values/quality flags. 

Also includes scripts to download and prepare the data in the correct format, as well as a test that checks these scripts on a tiny version of the dataset.

Related to https://github.com/AstroPile/FlatironMeeting2024/issues/12 and https://github.com/AstroPile/FlatironMeeting2024/issues/19.


al-jshen:
- currently doesn't include a download script for the main data file, I am not sure if there is a nice way to do the download but you can use a web interface [here](https://flathub.flatironinstitute.org/gaiadr3) and it has been slightly modified by me but it is on ceph on rusty. also not sure whether we will eventually like to include all of Gaia (i.e., both more rows and more columns)
- might want to think about including some default transform function to turn the spectral coefficients into wavelength/flux format, although the interface to do this with GaiaXPy is quite clunky

al-jshen:
For reference the data format looks something like this:
Screenshot 2024-03-27 at 4 01 25 PM

henrysky:
Can these columns be added for the following reason? `ruwe` for general quality cut; `nu_eff_used_in_astrometry`, `pseudocolour`, `ect_lat`, `astrometric_params_solved` for Gaia zero-point correction (whose are the columns required by `gaiadr3-zeropoint`); `rv_template_teff`, `grvs_mag` for radial velocity correction (see: https://ui.adsabs.harvard.edu/abs/2023A%26A...674A...7B/abstract, https://ui.adsabs.harvard.edu/abs/2023A%26A...674A...5K/abstract).

maja-jablonska:
We could think about adding a selection function accordng to the recommended cuts, e.g. ruwe<1.4, 5 params solved etc.

al-jshen:
@henrysky `ecl_lon` and `ecl_lat` added into the `astrometry` category. Also added a separate `corrections` category with all the info you requested. `ruwe` has been added to a `flags` category. I also separated out radial velocity info.

al-jshen:
> We could think about adding a selection function according to the recommended cuts, e.g., ruwe<1.4, 5 params solved etc. Maybe even apply some corrections ourselves - the end user shouldn't be bothered by diving into some Gaia's caveats

I'm not so sure that it would be good to impose particular cuts given that we expect this dataset to be used for a broad variety of potential downstream tasks because the selections/cuts that one makes is very dependent on the task. For example selecting `ruwe < 1.4` (probably good for most people) could cut out unresolved binaries, maybe someone out there is interested in. Some very broad cuts might be a good idea, but like you said we'd have to think very carefully about them.

al-jshen:
Someone could think about adding globus download to create the parent sample. The raw DR3 data are available at Flatiron here https://app.globus.org/file-manager/collections/90f5713a-e74d-11ec-9bd2-2d2219dcc1fa/overview. At the moment the parent sample is a massive HDF5 I built (and the `build_parent_sample.py` script just healpix-ifies it), but that could actually be fairly easily constructed from the files available here.

maja-jablonska:
> > We could think about adding a selection function according to the recommended cuts, e.g., ruwe<1.4, 5 params solved etc. Maybe even apply some corrections ourselves - the end user shouldn't be bothered by diving into some Gaia's caveats
> 
> I'm not so sure that it would be good to impose particular cuts given that we expect this dataset to be used for a broad variety of potential downstream tasks because the selections/cuts that one makes is very dependent on the task. For example selecting `ruwe < 1.4` (probably good for most people) could cut out unresolved binaries, maybe someone out there is interested in. Some very broad cuts might be a good idea, but like you said we'd have to think very carefully about them.

Yes, maybe ```ruwe``` is not the best example. But sometimes you can filter sources with poor astrometric solution.

scripts/gaia/README.md
```diff
@@ -0,0 +1,40 @@
...
+## Astrometry
+
+This includes:
+
+- `ra` and `dec` in degrees
+- `parallax` in mas
+- `pmra` and `pmdec` in mas/yr
+- `ra_error`, `dec_error`, `parallax_error`, `pmra_error`, and `pmdec_error` in mas
+- `ra_dec_corr`, `ra_parallax_corr`, `ra_pmra_corr`, `ra_pmdec_corr`, `dec_parallax_corr`, `dec_pmra_corr`, `dec_pmdec_corr`, `parallax_pmra_corr`, `parallax_pmdec_corr`, and `pmra_pmdec_corr` as correlation coefficients
+
```
maja-jablonska, line 16: Let's add a short explanation on RV here - I can see it's added in the dataset 😄

maja-jablonska (approved):
Looks all good in my opinion! 👏

scripts/gaia/README.md
al-jshen: README expanded

al-jshen:
Added some scripts to download and prepare the appropriate data files to build the parent sample, and then prepare everything for the HF dataset.

Added a `test.sh` script which downloads a small amount of data, prepares it, and loads it.

PR #32, opened 2024-03-28
[Infrastructure] Add master catalogue

ConnorStoneAstro:
Solves number 27 in the meeting issues


ConnorStoneAstro:
This doesn't actually work yet, but this PR is to help build it

ConnorStoneAstro:
The master catalogue now seems to build properly by extracting common fields from all AstroPile datasets. You can try it out if you have the AstroPile tiny dataset using this little script:

``` python
from datasets import load_dataset_builder
from astropile.build_master_cat import build_catalogue
import numpy as np

hsc_builder = load_dataset_builder("content/data/AstroPile/hsc", trust_remote_code=True)
desi_builder = load_dataset_builder("content/data/AstroPile/desi", trust_remote_code=True)
sdss_builder = load_dataset_builder("content/data/AstroPile/sdss", trust_remote_code=True)

master_cat = build_catalogue([hsc_builder, desi_builder, sdss_builder], ["HSC", "DESI", "SDSS"])
# print the whole table
print(master_cat)
print(np.sum(master_cat["HSC"]), np.sum(master_cat["DESI"]), np.sum(master_cat["SDSS"]))
# print where HSC and DESI have matches
print(master_cat[(master_cat["HSC"].values & master_cat["DESI"].values)])
```

ConnorStoneAstro:
@segasai

astropile/build_master_cat.py
```diff
@@ -0,0 +1,106 @@
...
+    master_cat : pd.DataFrame
+        The master catalogue containing the combined information from all the
+        input catalogues.
+    """
+
+    # Set the columns for the master catalogue
+    master_cat = pd.DataFrame(
+        columns=["ra", "dec", "healpix"] + names + [f"{name}_idx" for name in names]
+    )
+
```
lhparker1, line 50: Maybe you can throw some specific error if the len(names) != len(cats)

astropile/build_master_cat.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+
+        # Append the new rows to the master catalogue
+        master_cat = pd.concat([master_cat, append_cat], ignore_index=True)
+
+    # Convert the columns to the correct data types
+    master_cat["ra"] = master_cat["ra"].astype(float)
+    master_cat["dec"] = master_cat["dec"].astype(float)
+    master_cat["healpix"] = master_cat["healpix"].astype(int)
+    for name in names:
```
lhparker1, line 102: correct me if I'm wrong, but you already specify the bool and int type above (see lines 79-80 and 83-84, 84 you might need to specify dtype=int), probably don't need to do it again right?

lhparker1: Also, you can probably specify the dtype for ra dec and healpix above as well and get rid of this whole code chunk?

lhparker1 (commented):
See specific comments. As a general comment (also asked in slack), I'm curious if this should all live in utils, given that it is a similar thing to cross_match_datasets which does already live in utils.py. What are your thoughts? Can also ask @EiffL...

lhparker1 (commented):
Looks good!

lhparker1 (approved):

PR #38, opened 2024-03-28
[WIP] Adding GitHub Action to automatically test creation of datasets

EiffL:
This PR aims to add a simple testing mechanism that will try to download, format, and then load, a tiny version of the datasets.

This is not full proof of course when the full dataset will have to run, but it will allow us to catch a lot of the basic downloading and formatting issues.


EiffL (commented):
LGTM!

PR #44, opened 2024-04-04
dont run tiny_dset workflow when markdown files are changed

al-jshen:
don't run tiny_dset test when changing, for example, documentation, which doesn't affect the dataset prep/loading


EiffL (approved):
LGTM!

EiffL:
Awesome! Thanks Jeff!

PR #47, opened 2024-04-16
gz10 dataset

lhparker1:
Expansion to include gz10 dataset as a huggingface dataset that can be cross-matched with "data" datasets for tasks.

### DIFF
--- a/scripts/gz10/README.md
+++ b/scripts/gz10/README.md
@@ -28,19 +28,19 @@ This script will create a directory structure with the dataset split into multip
 
 ## Loading the Dataset with Hugging Face `datasets`
 
-To use the dataset in your projects with different configurations, specify the appropriate builder configuration when loading the dataset:
+To use the dataset in your projects with different configurations, specify one of the following appropriate builder configuration when loading the dataset. The dataset script `gz10.py` as the files processed with the `build_paranet_sample.py` script should all lie in the same current folder.
 
 ```python
 from datasets import load_dataset
 
 # For dataset with healpix indices (no images)
-dataset = load_dataset('path_to_dataset_script', name="gz10_with_healpix")
+dataset = load_dataset('gz10.py', name="gz10_with_healpix", trust_remote_code=True)
 
 # For the entire dataset from a single HDF5 file
-dataset = load_dataset('path_to_dataset_script', name="gz10_images")
+dataset = load_dataset('gz10.py', name="gz10_images", trust_remote_code=True)
 
 # For dataset with healpix indices and images
-dataset = load_dataset('path_to_dataset_script', name="gz10_with_healpix_with_images")
+dataset = load_dataset('gz10.py', name="gz10_with_healpix_with_images", trust_remote_code=True)
 ```
 
 ### Configurations
--- a/scripts/gz10/README.md
+++ b/scripts/gz10/README.md
@@ -1,6 +1,6 @@
 # GZ10 Galaxy Catalog Dataset
 
-The GZ10 catalog, provided by Leung et al. (2018), features 17,736 galaxies with labels from the Galaxy Zoo 2 project. 
+The Galaxy10 DECaLS dataset. This dataset combines images from the DESI Legacy Imaging Surveys (Dey A. et al., 2019) assembled by Walmsley et al., 2021 in the Galaxy Zoo DECaLS Campaign along with combined dataset classification labels from the Galaxy Zoo DECaLS campaign, the original Galaxy Zoo Campaign (Lintott et al. 2008), and the Galaxy Zoo DR2 release (Dey A. et al., 2019). Of these images and labels, 18k were selected in 10 broad classes. 
 
 ## Getting Started
 
--- a/scripts/gz10/README.md
+++ b/scripts/gz10/README.md
@@ -33,14 +33,15 @@ To use the dataset in your projects with different configurations, specify one o
 ```python
 from datasets import load_dataset
 
-# For dataset with healpix indices (no images)
-dataset = load_dataset('gz10.py', name="gz10_with_healpix", trust_remote_code=True)
+# For galaxy zoo 10 labels with no uint8 (rgb) images.
+dataset = load_dataset('gz10.py', name="gz10", trust_remote_code=True)
 
-# For the entire dataset from a single HDF5 file
+# For galaxy zoo 10 labels with uint8 (rgb) images.
+dataset = load_dataset('gz10.py', name="gz10_rgb_images", trust_remote_code=True)
+
+# For the entire dataset from a single HDF5 file. This does not support cross-matching.
 dataset = load_dataset('gz10.py', name="gz10_images", trust_remote_code=True)
 
-# For dataset with healpix indices and images
-dataset = load_dataset('gz10.py', name="gz10_with_healpix_with_images", trust_remote_code=True)
 ```
 
 ### Configurations
--- a/scripts/gz10/README.md
+++ b/scripts/gz10/README.md
@@ -6,7 +6,7 @@ The Galaxy10 DECaLS dataset. This dataset combines images from the DESI Legacy I
 
 ### Downloading the Dataset
 
-To download the GZ10 catalog, use the following `wget` command:
+To download the Galaxy10 DECaLS catalog, use the following `wget` command:
 
 ```bash
 wget https://www.astro.utoronto.ca/~hleung/shared/Galaxy10/Galaxy10_DECals.h5
--- a/scripts/gz10/README.md
+++ b/scripts/gz10/README.md
@@ -38,9 +38,6 @@ dataset = load_dataset('gz10.py', name="gz10", trust_remote_code=True)
 
 # For galaxy zoo 10 labels with uint8 (rgb) images.
 dataset = load_dataset('gz10.py', name="gz10_rgb_images", trust_remote_code=True)
-
-# For the entire dataset from a single HDF5 file. This does not support cross-matching.
-dataset = load_dataset('gz10.py', name="gz10_images", trust_remote_code=True)
 ```
 
 ### Configurations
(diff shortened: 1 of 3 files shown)


scripts/gz10/build_parent_sample.py
```diff
@@ -0,0 +1,60 @@
+import h5py
+import healpy as hp
+import numpy as np
+import os
+from tqdm import tqdm
+import argparse
+from multiprocessing import Pool
+
+def process_index(args):
```
mwalmsley, line 9: Generally this code needs docstrings, but to prioritise, this function is probably the most important to explain

scripts/gz10/build_parent_sample.py
```diff
@@ -0,0 +1,60 @@
...
+        list(tqdm(pool.imap(process_index, pool_args), total=len(pool_args), desc="Processing HEALPix indices"))
+
+def main(args):
+    output_dir = os.path.join(args.output_dir, 'datafiles')
+    if not os.path.exists(output_dir):
+        os.makedirs(output_dir)
+    save_in_standard_format(args.input_path, output_dir)
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description="Convert an HDF5 file to the standard format for the HEALPix-based dataset")
```
mwalmsley, line 56: Not specific to this PR, but, is there a specification written somewhere for what that standard format is?
(within astropile I mean, not HEALPix in general)

scripts/gz10/build_parent_sample.py
```diff
@@ -0,0 +1,60 @@
...
+
+        ra = file['ra'][:]
+        dec = file['dec'][:]
+        healpix_indices = hp.ang2pix(16, ra, dec, lonlat=True, nest=True)
+        unique_indices = np.unique(healpix_indices)
+    
+    print(f"Found {len(unique_indices)} unique HEALPix indices")
+
+    object_ids = np.arange(len(ra))
+    pool_args = [(input_path, output_dir, index, healpix_indices == index, object_ids)
```
mwalmsley, line 43: Here and in general, inline comments would be greatly appreciated

e.g.
For each healpix index, save all the rows (galaxies) from the original h5py file that have RA/Dec under a subfolder named with that index

scripts/gz10/build_parent_sample.py
```diff
@@ -0,0 +1,60 @@
...
+    with h5py.File(input_path, 'r') as file:
+        grouped_data = {key: file[key][mask] for key in file.keys()}
+    
+    grouped_object_ids = object_ids[mask]
+    output_subdir = os.path.join(output_dir, f'healpix={index}')
+
+    if not os.path.exists(output_subdir):
+        os.makedirs(output_subdir)
+
+    output_path = os.path.join(output_subdir, '001-of-001.h5')
```
mwalmsley, line 21: Is this an astropile convention?

scripts/gz10/build_parent_sample.py
```diff
@@ -0,0 +1,60 @@
...
+    
+    with h5py.File(output_path, 'w') as output_file:
+        for key in grouped_data:
+            output_file.create_dataset(key, data=grouped_data[key])
+        output_file.create_dataset('object_id', data=grouped_object_ids)
+        output_file.create_dataset('healpix', data=np.full(grouped_object_ids.shape, index))
+
+def save_in_standard_format(input_path, output_dir):
+    with h5py.File(input_path, 'r') as file:
+        required_keys = ['ans', 'dec', 'images', 'pxscale', 'ra', 'redshift']
```
mwalmsley, line 31: What is `ans`? Why does `pxscale` have no i? Generally, are these keys matching some astropile convention?

scripts/gz10/README.md
```diff
@@ -0,0 +1,50 @@
...
+The GZ10 catalog, provided by Leung et al. (2018), features 17,736 galaxies with labels from the Galaxy Zoo 2 project. 
+
+## Getting Started
+
+### Downloading the Dataset
+
+To download the GZ10 catalog, use the following `wget` command:
+
+```bash
+wget https://www.astro.utoronto.ca/~hleung/shared/Galaxy10/Galaxy10_DECals.h5
```
mwalmsley, line 12: This is the DECaLS version, while you elsewhere say it is the original GZ10 SDSS version

scripts/gz10/README.md
```diff
@@ -0,0 +1,50 @@
+# GZ10 Galaxy Catalog Dataset
+
+The GZ10 catalog, provided by Leung et al. (2018), features 17,736 galaxies with labels from the Galaxy Zoo 2 project.
```
mwalmsley, line 3: Please cite the original labels as well (Willet 2013 if you mean the GZ2 version, though I'm not sure if that's the version you mean)

scripts/gz10/gz10.py
```diff
@@ -0,0 +1,145 @@
...
+import os
+import datasets
+from datasets.data_files import DataFilesPatternsDict
+import itertools
+import h5py
+import numpy as np
+
+# TODO: Add BibTeX citation
+# Find for instance the citation on arxiv or on the dataset repo/website
+_CITATION = """\
```
mwalmsley, line 29: TODO

scripts/gz10/gz10.py
```diff
@@ -0,0 +1,145 @@
...
+# TODO: Add description of the dataset here
+# You can copy an official description
+_DESCRIPTION = """\
+The GZ10 catalog from Leung et al. (2018) is a dataset of 17,736 galaxies with labels from the Galaxy Zoo 2 project.
+The catalog includes the following features for each galaxy: right ascension, declination, redshift, and a label from the Galaxy Zoo 2 project.
+https://astronn.readthedocs.io/en/latest/galaxy10.html
+"""
+
+# TODO: Add a link to an official homepage for the dataset here
+_HOMEPAGE = ""
```
mwalmsley, line 47: TODO

scripts/gz10/gz10.py
```diff
@@ -0,0 +1,145 @@
...
+
+# TODO: Add the licence for the dataset here if you can find it
+_LICENSE = ""
```
mwalmsley, line 50: TODO

scripts/gz10/gz10.py
```diff
@@ -0,0 +1,145 @@
...
+
+_VERSION = "0.0.1"
+
+
+# TODO: Name of the dataset usually matches the script name with CamelCase instead of snake_case
+class GZ10(datasets.GeneratorBasedBuilder):
+    """TODO: Short description of my dataset."""
+
+    VERSION = datasets.Version("1.1.0")
```
mwalmsley, line 59: Different version to above?

scripts/gz10/gz10.py
```diff
@@ -0,0 +1,145 @@
...
+            {
+                "gz10_label": datasets.Value("int32"),
+                "ra": datasets.Value("float32"),
+                "dec": datasets.Value("float32"),
+                "object_id": datasets.Value("string"),
+            }
+        )
+
+        if self.config.name == "gz10_images" or self.config.name == "gz10_with_healpix_with_images":
+            features["images"] = datasets.Array3D(shape=(self._image_size, self._image_size, 3), dtype="uint8")
```
mwalmsley, line 93: Is this different to the other astropile datasets? I get a little lost but I thought people were keen on recording the flux rather than the processed uint8 images.

scripts/gz10/gz10.py
```diff
@@ -0,0 +1,145 @@
...
+
+                for k in keys:
+                    # Extract the indices of requested ids in the catalog 
+                    i = sort_index[np.searchsorted(sorted_ids, k)]
+
+                    example = {
+                        "gz10_label": data["ans"][i].astype(np.int32),
+                        "dec": data["dec"][i].astype(np.float32),
+                        "ra": data["ra"][i].astype(np.float32),
+                    }
```
mwalmsley, line 136: You required redshift in the HDF5 but don't read it out here. Not a problem necessarily, just curious.

scripts/gz10/build_parent_sample.py
```diff
@@ -0,0 +1,60 @@
...
+            raise ValueError("Missing one or more required datasets in the HDF5 file")
+
+        ra = file['ra'][:]
+        dec = file['dec'][:]
+        healpix_indices = hp.ang2pix(16, ra, dec, lonlat=True, nest=True)
+        unique_indices = np.unique(healpix_indices)
+    
+    print(f"Found {len(unique_indices)} unique HEALPix indices")
+
+    object_ids = np.arange(len(ra))
```
mwalmsley, line 42: Document in readme that `object_id` is arbitrary and not supposed to cross-match anything?
(more generally, document the possible HF keys in the readme)

Would it be sensible to use an astronomical id from SDSS or DESI-LS? Optional.

mwalmsley (changes requested):
Hi Liam. Nice work! 

The code is very clever/minimal/implicit. This is nice engineering but makes it quite hard to follow what each piece is doing. I think it could really benefit from some docstrings and inline comments throughout.

Similarly, there are many places in the code which seem to aim to conform to some astropile standards (I assume) but I'm unsure what those standards are and so am unsure how to review. Maybe link to where they are written down?

I think my only major comment is that you should check which version of GZ10 you are intending to use (SDSS vs DECaLS). Currently you say SDSS and use DECaLS, I think. I have left many minor comments inline.

mwalmsley:
I am going to remove the GZ10 dataset from #23 as it is superceded by this PR

mwalmsley:
I've started adapting the decals baseline to use this dataset instead of my earlier one. I ran into a few practical issues.

Major notes:
- The original h5 doesn't include `object_id` so can't be loaded `object 'object_id' doesn't exist`
- I notice this version dropped the whole `[images_by_band][array]` convention and returns uint8 images? (My version used the fits/flux)
- Relatedly, the `images` key is channels-last uint8, but dataset wrapper assumes channels-first (when taking a mean) and I'm not sure if people are intended to manage their own conversion to pytorch format (dataset.to_pytorch does not swap channels or convert uint8)

Minor notes:
- You need trust_remote_code=True, even when pointing to a local path. Update readme.
- wget command must place h5 in same directory as the script. Add -O flag (or at least update readme instructions)

mwalmsley:
Thanks for the updates Liam. 
- The dataset features now don't match the example keys. On my version I am still seeing 'images' rather than 'rgb_images' and I assume this may be why (otherwise might be HF caching?). In turn, that means the yielded examples still have the wrong channel position for PyTorch and so normalisation and training fail.

mwalmsley (approved):

PR #50, opened 2024-04-18
BTSbot

tom-hehir:
# Adds BTSbot training data to AstroPile

Collaboration with @mattgrayling @tom-hehir @benboyd97 @erinhay 

The Zwicky Transient Facility (ZTF)'s Bright Transient Survey (BTS) is the largest spectroscopic supernova survey ever conducted, and has been surveying the entire nighttime Northern sky every 2-3 nights in two filters and acquiring spectra of (almost) every time-varying, nonmoving object brighter than 18.5 mag that is not a Galactic source or an AGN, providing a complete and unbiased view of the dynamic optical sky.

The BTSbot training set is a multi-modal data set containing transient metadata and associated postage stamp images of the transients themselves. The data set comprises 409,107 transient alerts, and the goal of the BTSbot model was to determine whether each alert corresponded to a real transient that passed BTS's selection requirements meaning that it needed spectroscopic follow-up for a classification.

## Links for reference

- [paper](https://arxiv.org/abs/2401.15167)
- [data](https://zenodo.org/records/10839691)
- [model repo](https://github.com/nabeelre/BTSbot)

## HuggingFace Dataset Format

The HF dataset format resembles that of other image datasets e.g. HSC, except the image arrays comprise 3 different 'views' of the target: a reference image (no transient present), a science image (transient present), and the difference between these images. We present these views in the same way that different bands are presented in other image datasets. Concretely, the image data is structured as:
```
{
'view': ['science', 'reference', 'difference'],
'array': array with shape [63, 63, 3]
}
```
As usual, the dataloader returns a dictionary where the image data is accessible via the key 'image'. This dictionary also includes all metadata from the BTSbot dataset e.g. `ra`, `dec`, `is_SN`, etc. as well as `healpix`.


tom-hehir:
**Ready for review**
- PR has all the required functionality and is passing tests, except for the known issue with HSC which needs to be resolved externally to this PR.
- One small issue: when calling the dataloader we get the warning `Repo card metadata block was not found. Setting CardData to empty.` I don't understand this warning, but it doesn't seem to affect functionality of the dataloader and I expect it will be a quick fix.

nabeelre:
More up to date paper for this dataset: [https://arxiv.org/abs/2401.15167](https://arxiv.org/abs/2401.15167)

tom-hehir:
> More up to date paper for this dataset: https://arxiv.org/abs/2401.15167

Thanks for letting us know about the latest paper @nabeelre -- I will update the files to point this paper. And congratulations on your results with BTSbot, very cool !!

tom-hehir:
> **Ready for review**
> 
> * PR has all the required functionality and is passing tests, except for the known issue with HSC which needs to be resolved externally to this PR.
> * One small issue: when calling the dataloader we get the warning `Repo card metadata block was not found. Setting CardData to empty.` I don't understand this warning, but it doesn't seem to affect functionality of the dataloader and I expect it will be a quick fix.

To eliminate the `Repo card metadata block was not found. Setting CardData to empty` warning, you need to call `datasets-cli test path/to/<your-dataset-loading-script> --save_info --all_configs` which will prepend some information about the dataset to the README.md in the dataset directory. However this causes testing to fail, because testing uses a limited version of the dataset with different repo card metadata to the full version, which causes `load_dataset` to raise an error. I don't see an easy way around this, and the warning does not affect functionality in any important ways (it just prevents a loading bar from showing while examples are generated), so I think it's okay to proceed with the warning.

tom-hehir:
@mb010 Just tagging you so you see this (not sure if adding you as a reviewer sends a notification).

scripts/btsbot/README.md
```diff
@@ -0,0 +1,4 @@
+# TODO:
+- [ ] Add image scale (1.01" per pixel for ZTF)
+- [ ] Obtain PSFs and incorporate into dataset
+- [ ] Implement same train/val/test splits as BTSbot dataset
```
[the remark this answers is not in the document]
tom-hehir: Thanks for looking over things!
Short answer: no, but I think it might be okay to leave these as to-dos for v0 (and can complete them in a separate PR in future).
2. So eventually I would like to get the PSF files for all the images (it will be a fair bit of work and isn't essential information for the intended use of this dataset), but I checked with Francois and apparently it should be okay to omit this info for the version 0 of AstroPile.
1. I found out that the image scale of ZTF images is 1.01" per pixel. I could include this easily now, but was going to do it at the same time as (2).
3. For now I was going to agglomerate all of train/val/test into train, and the user can define their own splits for now.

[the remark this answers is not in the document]
mb010: Omit psf file is fine for now.
Include as float. 
Split as before if possible. 👍

scripts/btsbot/_make_tiny_download.py
```diff
@@ -0,0 +1,60 @@
+import argparse
+import numpy as np
+import pandas as pd
+import os
+import shutil
+
```
mb010, line 6: Write a docstring explaining what this file is used for and linking to the tiny data that was created with it.

scripts/btsbot/btsbot.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+class BTSbot(datasets.GeneratorBasedBuilder):
+    """BTSbot training set"""
+
+    VERSION = _VERSION
+    BUILDER_CONFIGS = [
+        datasets.BuilderConfig(name="BTSbot_training_set", 
+                               version=VERSION,
+                               data_files=DataFilesPatternsDict.from_patterns({'train': ['./data/healpix=*/*.hdf5']}),
```
mb010, line 152: This is where data splits are defined @mattgrayling.

scripts/btsbot/_get_features.py
```diff
@@ -0,0 +1,37 @@
+
+import h5py
+
```
mb010, line 3: Doc string to explain what this file is.

scripts/btsbot/README.md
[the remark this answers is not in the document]
mb010: Include wavelength / band feature.

scripts/btsbot/download_data.py
```diff
@@ -0,0 +1,57 @@
...
+
+        # Remove tar.gz file
+        os.remove(os.path.join(args.destination_path, file_name))
+
+    # Print a success message if the file is downloaded successfully
+    print(f"Files downloaded successfully to {args.destination_path}")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Download BTSbot data, optionally downloading a small subset for testing.")
```
mb010, line 50: "Download BTSbot data, optionally downloading a small subset for testing." -> "Download BTSbot data, optionally download a small subset for testing instead."

mb010 (changes requested):
I left some comments to be addressed. A few small changes and the data splitting to address. Well done.

scripts/btsbot/btsbot.py
tom-hehir: @mattgrayling I have modified this part of `btsbot.py` to support train/val/test splits. It now expects 3 separate files per healpix directory, with `train_001-of-001.hdf5` storing training examples only, `val_001-of-001.hdf5` storing val examples only, and `test_001-of-001.hdf5` storing test examples only.
NOTE: the lines I've added are currently commented out to avoid breaking things -- they need to be uncommented (and old line deleted) once saving has been made consistent with this format.

tom-hehir:
@mattgrayling looking at the paper, it seems like the data could be a mix of g and r band, but I don't see a field for band in the metadata. I checked which of the metadata keys are binary valued, and these are the options:
```
label
fid
exptime
is_SN
near_threshold
is_rise
```
Perhaps `fid` is the band (I don't know what it stands for), but its values are 1/2, so we'd need to establish the correspondence with r/g.

nabeelre:
unsubscribing from this to stop the notifications but feel free to tag me if you have any questions about the BTSbot data

[This page](https://zwickytransientfacility.github.io/ztf-avro-alert/schema.html) provides descriptions with units for the metadata fields which originate from ZTF alerts. Some others (`is_SN`, `near_threshold`, `is_rise`) are custom defined by me for testing purposes, but, if you're very curious, most will be [defined in the BTSbot repo](https://github.com/nabeelre/BTSbot/blob/main/alert_utils.py)

tom-hehir:
> unsubscribing from this to stop the notifications but feel free to tag me if you have any questions about the BTSbot data
> 
> [This page](https://zwickytransientfacility.github.io/ztf-avro-alert/schema.html) provides descriptions with units for the metadata fields which originate from ZTF alerts. Some others (`is_SN`, `near_threshold`, `is_rise`) are custom defined by me for testing purposes, but, if you're very curious, most will be [defined in the BTSbot repo](https://github.com/nabeelre/BTSbot/blob/main/alert_utils.py)

Thanks so much!!

scripts/btsbot/README.md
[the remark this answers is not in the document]
tom-hehir: Adding band and scale now resolved.
Adding train/val/test splits is resolved on the data loader end, but not the hdf5 file saving end.

scripts/btsbot/btsbot.py
tom-hehir: Oh and if there are no train/val/test examples for a given healpix, then it's fine to omit the relevant files.

scripts/btsbot/README.md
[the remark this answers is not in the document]
tom-hehir: All complete

tom-hehir:
@mb010 All improvements from your review have now been implemented, tests are passing , and no conflicts with main -- should be ready to merge after approving review.

mb010 (approved):

PR #51, opened 2024-04-18
5 Type Ia Supernova Datasets From PantheonPlusSH0ES

benboyd97:
# Adds 5 Type Ia Supernova Datasets to Astropile From PantheonPlusSH0ES
Collaboration with @tom-hehir @mb010 @erinhay @David-Chemaly @ado8 @mattgrayling and others.

Adds five multiband time-series Type Ia supernova datasets from DES Y3, Swift, PS1, Foundation DR1 and SNLS gathered by PantheonPlusSH0ES. This code is almost identical to the yse code in #46. 

## Links for reference:
[PantheonPlusSH0ES Data](https://github.com/PantheonPlusSH0ES/DataRelease/tree/main/Pantheon%2B_Data/1_DATA/photometry) 
- [PantheonPlusSH0ES Paper](https://iopscience.iop.org/article/10.3847/1538-4357/ac8b7a/pdf)
- [DES Y3 Paper](https://iopscience.iop.org/article/10.3847/1538-4357/ab06c1/pdf)
- [Swift Paper](https://arxiv.org/pdf/1407.3808.pdf)
- [PS1 Paper](https://iopscience.iop.org/article/10.3847/1538-4357/aab9bb/pdf)
- [Foundation DR1 Paper](https://iopscience.iop.org/article/10.3847/1538-4357/ab2bec/pdf)
- [SNLS Paper](https://www.aanda.org/articles/aa/pdf/2010/15/aa14468-10.pdf)

## HuggingFace Dataset Format
Each example contains:

  - `object_id`: Dataset specific identifier
  - `obj_type`: Spectral class
  - `ra`, `dec`: right ascension, declination
  - `redshift`: redshift of target. : As decided by ordered availability of: `['REDSHIFT_FINAL', 'REDSHIFT_CMB', 'REDSHIFT_HELIO']`
  - `host_log_mass`: host mass. As decided by ordered availability of: `['HOST_LOGMASS', 'HOSTGAL_LOGMASS']` 
  - `lightcurve`: object with:
                - `band`: band of observation, arr(string),
                - `time`: timestamp of observation in MJD, arr(float)
                - `flux`: flux value of observation, arr(float),
                - `flux_err`: flux uncertainty of observation, arr(float)


scripts/des_y3_sne_ia/download_data.py
```diff
@@ -0,0 +1,43 @@
...
+
+def main(args):
+
+    # Construct the URL to download 
+    url = 'https://raw.githubusercontent.com/PantheonPlusSH0ES/DataRelease/main/Pantheon%2B_Data/1_DATA/photometry/DES3YR_DES_COMBINED_TEXT/'
+    
+    #name of files
+    names=read_text_file('DES3YR_DES_LIST.txt').split('\n')[:-1]
+
+    if not os.path.exists(args.destination_path+'/des_y3_sne_ia/'):
```
helenqu, line 31: nit: more canonical to use `os.path.join(orig_path, subdir_name)` or `from pathlib import Path; Path(orig_path) / "subdir_name"` to create new paths

scripts/des_y3_sne_ia/download_data.py
```diff
@@ -0,0 +1,43 @@
...
+        os.makedirs(args.destination_path+'/des_y3_sne_ia/')
+        
+    for file in names:
+        download_text_file(url+file,args.destination_path+'/des_y3_sne_ia/'+file)
+    
+ 
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Transfer data from YSE DR1 to user-provided destination folder.")
```
helenqu, line 40: nit: wrong description

scripts/des_y3_sne_ia/des_y3_sne_ia.py
```diff
@@ -0,0 +1,190 @@
...
+            if isinstance(files, str):
+                files = [files]
+            # Use `dl_manager.iter_files` to skip hidden files in an extracted archive
+            files = [dl_manager.iter_files(file) for file in files]
+            splits.append(
+                datasets.SplitGenerator(name=split_name, gen_kwargs={"files": files})
+            )
+        return splits
+
+    def _generate_examples(self, files):
```
helenqu, line 151: i've seen many `generate_examples` with an optional `object_ids` argument, e.g. https://github.com/AstroPile/AstroPile_prototype/pull/46/files#diff-8752406b3065c128d74fe897f90a753297385f758618519efdecd8c011db8be7R227

scripts/des_y3_sne_ia/des_y3_sne_ia.py
```diff
@@ -0,0 +1,190 @@
...
+
+                # Preparing an index for fast searching through the catalog
+                sort_index = np.argsort(data["object_id"][()])  # Accessing the scalar index
+                sorted_ids = [data["object_id"][()]]  # Ensure this is a list of one element
+
+                for k in keys:
+                    # Extract the indices of requested ids in the catalog
+                    i = sort_index[np.searchsorted(sorted_ids, k)]
+                    # Parse data
+                    idxs = np.arange(0, data["flux"].shape[0])
```
helenqu, line 166: i don't see `i` used anywhere in the `for k in keys` loop so i'm kind of confused since `i` is the object id we're trying to load in to `example`?

scripts/des_y3_sne_ia/des_y3_sne_ia.py
```diff
@@ -0,0 +1,190 @@
...
+                        }
+                    }
+                    # Add remaining features
+                    for f in _FLOAT_FEATURES:
+                        example[f] = np.asarray(data[f]).astype("float32")
+                    for f in _STR_FEATURES:
+                        # Add band names shared across dataset to each sample.
+                        # I can't see a better way to do this.
+                        if f == "bands":
+                            example[f] = data[f][()].decode('utf-8')
```
helenqu, line 186: why does this need `decode('utf-8')` again? it seems like when you make the `bands` array you already did `decoe('utf-8')`?

helenqu (commented):
only checked the DES3YR files since we discussed at the last meeting that all the files are basically the same across surveys

scripts/des_y3_sne_ia/des_y3_sne_ia.py
tom-hehir: I think this is due to there being only one example per hdf5 file, whereas this code has been inherited from other datasets where there are multiple examples per hdf5 file (in which case the index is used to select the particular examples required from the file). It should therefore be safe to remove `sort_index`, `sorted_ids`, and `i`. In future, we may want to switch to storing multiple examples per file and restore these variables.

tom-hehir: Actually we need to do a bit more than this, otherwise object_ids, k, and keys become unused variables, and we aren't able to selectively return data based on the provided object_ids. A quick fix would be to just iterate over those files that are requested as object_ids (or all files if no object_ids provided).

scripts/des_y3_sne_ia/download_data.py
benboyd97: good spot! I've fixed this for all five surveys.

scripts/des_y3_sne_ia/download_data.py
benboyd97: Ah my mistake I should have changed this after experimentation. I changed the code and download_data.py for all five surveys now uses os.path.join

scripts/des_y3_sne_ia/des_y3_sne_ia.py
benboyd97: This has now been fixed for all five surveys matching @tom-hehir solution in https://github.com/AstroPile/AstroPile_prototype/blob/3d6da624b945b1c993d8447042a23e4453ccaac0/scripts/yse/yse.py

scripts/des_y3_sne_ia/des_y3_sne_ia.py
benboyd97: This has now been fixed for all five surveys matching @tom-hehir solution in https://github.com/AstroPile/AstroPile_prototype/blob/3d6da624b945b1c993d8447042a23e4453ccaac0/scripts/yse/yse.py

helenqu (approved):

PR #53, opened 2024-04-18
Adding 4 Harvard Center for Astronomy datasets.

ado8:
# Adds 4 supernova datasets to Astropile from Harvard's Center for Astronomy
Collaboration with @tom-hehir @mb010 @erinhay @David-Chemaly @benboyd97 @mattgrayling and others.
Includes CfA 3, CfA 4, CfA Stripped-envelope Core-collapse supernovae, and CfA Type II Supernovae datasets.
The download_data and build_parent_sample python files are shared between all datasets, requiring an argument of cfa3, cfa4, cfa3_4sh, or cfa_snII respectively.

## Links for reference:
[CfA Data Products](https://lweb.cfa.harvard.edu/supernova/) 
- [CfA3 paper](https://ui.adsabs.harvard.edu/abs/2009ApJ...700..331H/abstract)
- [CfA4 paper](https://ui.adsabs.harvard.edu/abs/2012ApJS..200...12H/abstract)
- [CfA Stripped-Envelope Core-collapse paper](https://ui.adsabs.harvard.edu/abs/2014ApJS..213...19B/abstract)
- [CfA Type II paper](https://ui.adsabs.harvard.edu/abs/2017ApJS..233....6H/abstract)

## HuggingFace Dataset Format
Each example contains:

  - `object_id`: Dataset specific identifier
  - `spec_class`: Spectral class
  - `ra`, `dec`: right ascension, declination
  - `lightcurve`: object with:
                - `band`: band of observation, arr(string),
                - `time`: timestamp of observation in MJD, arr(float)
                - `mag`: Magnitude (Standard system) value of observation, arr(float),
                - `mag_err`: Magnitude (Standard system) uncertainty of observation, arr(float)


scripts/.DS_Store
LTMeyer, line 1: Should this file be kept in the git history?

download_data.py
```diff
@@ -0,0 +1,43 @@
+import os
+import argparse
+import requests
+import tarfile
+import requests
+import os
```
LTMeyer, line 6: `os` is imported twice.

download_data.py
```diff
@@ -0,0 +1,43 @@
...
+        contents = file.read()
+        return contents
+    
+
+
+
+def main(args):
+
+    # Construct the URL to download 
+    url = 'https://raw.githubusercontent.com/djones1040/Foundation_DR1/master/Foundation_DR1/'
```
LTMeyer, line 26: Is this url gonna be permanently accessible? Is this the official source of data?

scripts/cfa/build_parent_sample.py
```diff
@@ -0,0 +1,315 @@
...
+
+survey_specific_logic = {
+    "cfa3": cfa3_bpf,
+    "cfa3_4sh": cfa_generic_bpf,
+    "cfa4": cfa_generic_bpf,
+    "cfa_snII": cfa_snII_bpf,
+}
+
+
+def main(args):
```
LTMeyer, line 202: Listing all arguments instead of grouping them into a single `args` would help in reviewing the logic of the function and detecting any unused or unprovided argument.

LTMeyer (commented):
Great work which mostly looks good to me. I have made few minor comments about the code.

ado8:
I'm not sure where the dot file crept in from, but I removed it.
I'm not great with managing git histories, but I tried getting rid of it via

> git filter-branch --invert-paths --path scripts/.DS_Store --force

but then had trouble pushing.
I deleted my local repo, cloned it again, and just removed the file, so
it's still in the history.

This also applies to the base directory download_data.py, which I think I
pulled in from the YSE or sne_ben branch for reference and forgot to prune.
It's also still in the history but probably shouldn't be.

RE: scripts/cfa/build_parent_sample.py
<https://urldefense.com/v3/__https://github.com/AstroPile/AstroPile_prototype/pull/53*discussion_r1584456526__;Iw!!PvDODwlR4mBZyAb0!Xynqr4Yp96QEEDOmYyo1_fKZ1w9XF6kzG7s_o9b0QA4vH1eqAl2CorlLXCJPXggJEqG16WBy0o_yCe_WlTjWUA$>
:
I unpacked the relevant args attributes into their own variables (e.g.
args.dataset -> dataset) in the main function and passed them explicitly to
the survey_specifc logic functions.
This results in functions that only work for one survey ('cfa3_bpf',
'cfa_snII_bpf') receiving an extra dataset argument so I just added
**kwargs to catch unused function arguments.

helenqu:
no additional comments from me, besides just going through and removing some commented out code (e.g. https://github.com/AstroPile/AstroPile_prototype/pull/53/files#diff-df76b783bac651fcd8c72ba302d550bf91d2e46cdfbba86cd778cbebd0bde681R57)

most importantly, make sure to resolve the merge conflict with the `.gitignore` file, it looks like you just added an extra line which is upsetting git.

ado8:
I think everything is ready to go for this and the CSP branch. Let me know if there's anything else I ought to do.

helenqu (approved):

PR #66, opened 2024-05-03
Fix Globus transfer for DESI dataset

### LINKED ISSUE #62: Provide More Details on Globus Download
I have never used globus before and I understand it is a utility for large data transfers.
I am trying to perform some tests with either the DECALS or DESI datasets. Both rely on an initial globus data transfer. Given the size of the files, I would like to download and process them on a cluster (typically Jean-Zay). However, the documentation on how to set up globus on a cluster and how to use it is quite limited.
Would it be worthy to extend the current documentation (e.g. https://github.com/AstroPile/AstroPile_prototype/blob/main/scripts/desi/README.md) to explain how to setup globus and download files on a generic machine that would include cluster? 

I posted a question on how to use globus on JZ on their gitter.

LTMeyer:
This PR removes some argument to the globus authentificator which were preventing from downloading the data. It also tries to provide more details in the documentation regarding data transfer through Globus.

Closes #62

### DIFF
--- a/scripts/desi/README.md
+++ b/scripts/desi/README.md
@@ -17,13 +17,19 @@ DESI Early Data Release applying only the following cuts:
 
 ### Downloading data through Globus
 
-The first step to data preparation is to download through globus all relevant data to a local machine. It is necessary either to use an institutional Globus endpoint or to create a personal one following the instructions here: https://www.globus.org/globus-connect-personal.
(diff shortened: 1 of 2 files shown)


EiffL (approved):
Great! Thank you so much Lucas, this looks good and seems to work

PR #69, opened 2024-05-16
ML baselines liam

lhparker1:


EiffL:
I've combined all baselines in one PR, reviewing this, I see a lot of potential for code synergies to avoid duplication of similar models or frameworks, so I'm going to attempt a little bit of reformatting and then we can merge :-)

EiffL:
@lhparker1 I made a proposal to streamline to structure of the benchmarks here: https://github.com/AstroPile/AstroPile_prototype/pull/76
Let me know what you think

scripts/gz10/gz10.py
```diff
@@ -124,6 +124,9 @@ def _split_generators(self, dl_manager):
 
     def _generate_examples(self, files, object_ids=None):
         """Yeilds examples from the dataset"""
+        # not sure why but I need to unpack files to avoid "train has no data" error
+        # Instruction "train" corresponds to no data!
+        files = list(files)
```
EiffL, line 129:
```suggestion
```

EiffL: this should no longer be necessary

PR #74, opened 2024-05-25
Implements a few efficiency tweaks to speed up file operations (including standardizing healpix_nside=16)

EiffL:
This PR makes a few changes to improve the efficiency of the dataloader. In particular:

- It removes the calls to `itertools.chain.from_iterable` in the generator function of the different datasets
- It changes the default healpix nside to 16 instead of 64, to reduce the number of files

These changes are quite critical to preserve efficient loading operations when dealing with a large number of files on a distributed filesystem.


EiffL:
All checks are passing, and this is a required update for #69 because we need a standardized healpix nside for cross-matching and building benchmarks.

PR #78, opened 2024-05-29
Spectra baselines for DESI and SDSS 

EiffL:
Here is some code for running a galaxy spectra -> redshift baseline for the DESI and SDSS spectra data. 

I've implemented a 1D ResNet-18 trained via the Huber loss for this, as in [GaSNet from Zhong+2023](https://arxiv.org/abs/2311.04146).

WIP: Still need to incorporate improvements from #76, and one-click training via Lightning.


Smith42:
Added ability to run this as a Lightning CLI, so now ready for review

PR #82, opened 2024-06-04
Photometric Classification Baseline using snmachine

ado8:
I made some python functions to translate YSE data into the format expected by an existing photo-classifier package called snmachine, perform some basic pre-processing, and wrap some snmachine functions.
I've updated the dataset requirements to require a fork of snmachine I made to fix some bugs.
Running
python snmachine_example.py
should produce some figures and an output_yse directory with some data products.
It's slow and a more elegant ML solution would have been preferable, but I think this demonstrates how the AstroPile datasets can be used with existing software.


dset-requirements.txt
```diff
@@ -11,6 +11,9 @@ git+https://github.com/changhoonhahn/provabgs.git
 #SN specific dependencies
 sncosmo
 
+# SN baseline specific dependencies
+git+https://github.com/ado8/snmachine
+
```
mb010, line 16: This should be in the readme of the baseline for reconstruction of the baseline. Not in the dataset specific requirements.
```suggestion
```

baselines/photo_class/README
```diff
@@ -0,0 +1,12 @@
+This baseline demonstrates how AstroPile data sets can be used with existing software, namely the snmachine package from LSST/DESC.
```
mb010, line 1:
```suggestion
This baseline requires a package which can be installed by: `pip install git+https://github.com/ado8/snmachine`

This baseline demonstrates how AstroPile data sets can be used with existing software, namely the snmachine package from LSST/DESC.
```

mb010 (approved):
minor fixes and then lgtm.

PR #90, opened 2024-06-12
Script to upload datasets to HuggingFace

EiffL:
This PR add a script that can be run to automatically upload a chunk of each dataset to the hub. 

It already kinda works, but does not upload the metadata like citation and licenses. I'm not sure if there is a way to do that, if not, we will have to manually track that information in two different places


EiffL:
I get the following errors:

Failed to upload chandra dataset
An error occurred while generating the dataset

Preparing yse dataset
Failed to upload yse dataset

Preparing manga dataset
Repo card metadata block was not found. Setting CardData to empty.
Failed to upload manga dataset

Preparing jwst dataset
Failed to upload jwst dataset
Config name is missing.
Please pick one among the available configs: ['primer-cosmos-east', 'ceers-full', 'ngdeep', 'primer-uds', 'gds', 'gdn', 'primer-cosmos-east-tiny', 'ceers-full-tiny', 'ngdeep-tiny', 'primer-uds-tiny', 'gds-tiny', 'gdn-tiny']
Example of usage:
        `load_dataset('jwst', 'primer-cosmos-east')`

LTMeyer:
# Issue with chandra dataset

From these lines (https://github.com/MultimodalUniverse/MultimodalUniverse/blob/main/scripts/chandra/chandra.py#L127-L137) and some debugging it appears that there are discrepancies between:
- The `object_ids` passed as parameters
- The number of object in the current file

As we loop over all the files of the dataset, the enumeration `j` can exceed the number of `object_ids` passed as parameters hence an error at l.130.

Due to the discrepancies and how the index of the data is retrieved (l.141), it can exceed the number of actually available data. Leading to an index error in l.147.

Because the logic is not yet totally clear to me of what we are trying to do, I don't see an obvious fix.

If `object_ids` are a set of ids we are looking in the whole dataset, the current logic should be adapted to something like:

```python
requested_objects = set(object_ids) # The object ids we are looking for
for file in files:
with h5py.File(file) as data:
    available_objects = set(data["object_ids"])
    for match_object in (requested_objects & available_objects):
        object_idx = list(available_objects).index(match_object)
        object_data = {field: data[field][object_idx] for field in data["field_names"]}
        yield match_object, object_data
```

LTMeyer:
# Issue with yse dataset

The issue was related to the data path. I fixed it to reflect what it was on rusty. However, it is still not consistent with default values in download script.

I've then uploaded the dataset to Hugging Face hub.

LTMeyer:
# Issue with jwst dataset

The following works. 
```python
set = datasets.load_dataset('/mnt/ceph/users/polymathic/MultimodalUniverse/jwst/jwst.py', "ceers-full" , trust_remote_code=True, split='train')
```

It seems not all configuration are currently available on rusty.

LTMeyer:
To get the Croissant json files for all the dataset of the AstroPile organization on HG run:

```bash
python get_croissant.py
```

LTMeyer (approved):
LGTM. I think we can merge this one. If there are still datasets to be fixed, we can do so in separated PRs.

PR #91, opened 2024-06-12
CfA homepage fixes

ado8:
Added homepage to cfa3.py. Trimmed homepages in cfa4.py cfa_snII.py t…o point at the pages holding the data files rather than the data files.


mb010 (approved):
All good.

PR #105, opened 2024-06-13
Update-readme

lhparker1:


README.md
```diff
@@ -1,4 +1,7 @@
 # Multimodal Universe: Enabling Large-Scale Machine Learning with 70TBs of Astronomical Scientific Data
+<!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
+[![All Contributors](https://img.shields.io/badge/all_contributors-28-orange.svg?style=flat-square)](#contributors-)
+<!-- ALL-CONTRIBUTORS-BADGE:END -->
 [![Testing tiny datasets](https://github.com/AstroPile/AstroPile_prototype/actions/workflows/tiny_dset_test.yml/badge.svg)](https://github.com/AstroPile/AstroPile_prototype/actions/workflows/tiny_dset_test.yml)
 [![All Contributors](https://img.shields.io/github/all-contributors/AstroPile/AstroPile_prototype?color=ee8449&style=flat-square)](#contributors)
```
EiffL, line 6:
```suggestion
```

EiffL:
Thank you Liam, it looks good to me, I think I would just remove `data` and just do `code` instead. 

I'm also not sure what `infra` covers?

lhparker1:
The idea was that code represented ppl who contributed to baselines, and data to data. Maybe it would be more clear to replace `data` with `code`, and do another emoji for baseline (does `example` make sense? not sure what this could be). 

`infra` is supposed to represent ppl who contributed to the general infrastructure design (like organizing HF structure, backend structure, croissant, etc.) -> thinking about you and Lucas for example

Smith42:
Might be a good idea to test if the All-Contributors bot still works after this README update @lhparker1?

EiffL:
Yeah thanks Liam, I think it would be clearer to have `example` for the baselines, and `code` for the rest of the contributions :-) Happy to approve after that

Issue #98, opened 2024-06-12
`object_id` attribute is missing from chandra data


LTMeyer:
The HDF5 should contain `object_id` field.

PR #106, opened 2024-06-14
Solves the issue with chandra missing object_id

EiffL:
This PR solves issue #98 by adding manually an object id at the catalog level. Also fixed bug with data loading which was a consequence of not having an object_id column in the dataset.


EiffL:
I double checked, works fine

PR #109, opened 2024-06-14
Bug fixes for manga dataset creation

### LINKED ISSUE #100: Issue with MaNGA when converting to arrow
@havok2063 I have the following issue with manga which is preventing me from uploading to HF.

The test works but the following doesn't:

```python
import datasets
from datasets import Dataset
from functools import partial

dset = datasets.load_dataset(f'/mnt/ceph/users/polymathic/MultimodalUniverse/manga', 
                    trust_remote_code=True, streaming=True, split='train')

dset = dset.take(5)

def gen_from_iterable_dataset(iterable_ds):
    yield from iterable_ds

ds = Dataset.from_generator(partial(gen_from_iterable_dataset, dset), 
                                    features=dset.features)
```

It doesnt work because the conversion to Arrow is not too happy about the way the data is structured. For spaxels at least I tracked it down to the fact that the flux field and such don't actually have the declared shape of a list of 2d arrays of size (1,N).

The same thing should happen with the tiny version of the dataset if you want to try locally

Could you take a look?

havok2063:
This PR closes #100.  It fixes a few bugs when creating the hugging face dataset.  The following issues were fixed:

- The spaxel 1d array data for `flux`, `ivar`, `mask`, `lsf`, and `lambda` were not in the proper shape of `[1, nwave]`
- The incorrect data extension was accessed for the `MASK` arrays.  This was causing some improper int/float conversions.
- The `spaxel_idx` had the incorrect integer type.  

This will require regeneration of the manga healpix files from the original dataset.  Once re-run, the code snippet from #100 should run.  Additional checks may be needed to ensure everything else in the hf dataset looks ok.

### DIFF
--- a/scripts/manga/build_parent_sample.py
+++ b/scripts/manga/build_parent_sample.py
@@ -210,7 +210,7 @@ def process_single_plateifu(args: tuple) -> dict:
         # pad rest with 0s
         flux = flux.reshape(nwave, nspaxels)
         ivar = np.pad(hdulist['IVAR'].data, pad_arr).reshape(nwave, nspaxels)
-        mask = np.pad(hdulist['FLUX'].data, pad_arr, constant_values=1024).reshape(nwave, nspaxels)
+        mask = np.pad(hdulist['MASK'].data, pad_arr, constant_values=1024).reshape(nwave, nspaxels)
         lsf = np.pad(hdulist['LSFPOST'].data, pad_arr).reshape(nwave, nspaxels)
 
         wave = hdulist['WAVE'].data.astype(np.float32)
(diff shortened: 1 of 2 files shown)


EiffL:
Thanks Brian, it looks very good. I'm currently doing a production run with these modifications and checking that everything works. Will merge once that's done.

havok2063:
@EiffL is there a progress update on this?  Did everything run ok?  Is there anything more I can do?

EiffL:
Hi Brian, I'm running into problems when trying to run the production code, my parallel processing script hangs and times out. I didn't have these issues before but I haven't yet figured out why this happens. A couple of healpix cells process just fine, and then it stops.

havok2063:
Hmm, that is strange.  Nothing has changed with the healpix generation code.  Could it be an issue with size of the healpix resolution (`_healpix_nside`)?   You already have all the data locally?  Does it always stop at the same data files each time you run it?

EiffL:
Yes, I have all the data locally. I suspect it's one of these that fails:
https://github.com/MultimodalUniverse/MultimodalUniverse/blob/2cca60efbea8b72c0edd6f512ad1e2e2e60d3c9e/scripts/manga/build_parent_sample.py#L298

But I haven't been able to get an error yet

EiffL:
I think I need to run without the multiprocessing to see what the issue is

mb010:
Superseded by #129.

PR #113, opened 2024-06-26
adding script to download legacysurvey

EiffL:
This small PR adds missing script to download the legacysurvey from NERSC through globus


EiffL:
@lhparker1 I've also added a small bit of additional code to regenerate the files that had an issue. The legacysurvey should all be fine now.

PR #115, opened 2024-07-02
[Draft] Add mask and catalog information to legacy survey

### LINKED ISSUE #110: Add mask datasets
# Description

We want to be able to retrieve the mask over celestial objects. Given an `object_id` we would like to have a segmentation map of all the objects in the cut out image. There are three possible sources for the mask information.

## Legacy Survey

The blob model provides an image we can threshold for segmentation. Preliminary tests suggest the `.jpg` image is better suited to obtain the mask. It remains to decide which threshold to use.

- [x] ~Decide the thresholding technique~
- [x] ~Decide how to export the mask (one-hot encoding, image, standalone-dataset etc.)~
- [x] Return the model image

For the legacy survey, we opt for returning directly the model image. It will allow users to set the threshold and generate the mask on the fly. It prevents from having to process the dataset each time we want to change the mask threshold value.

## Sweep Catalogs

From the sweep catalog, we want to retrieve the catalog itself of the objects in the cutout (besides obtaining the mask). 

## Galaxy Zoo 3D

We want to obtain the masks from this existing [dataset](https://academic.oup.com/mnras/article/507/3/3923/6347355?login=false).

LTMeyer:
Related to #100 

This PR adds to the legacysurvey dataset:
- blob model image;
- mask based on catalog information around the center object in the cutout;
- catalog information about the 20 brightest objects in the cutout. 

Closes #110


LTMeyer:
Close PR because superseded by #118.

PR #117, opened 2024-07-10
expand Gaia dataset to include full Gaia source tables and RVS spectra

al-jshen:
now includes:
- Gaia Source, full tables (1TB, 1.8B rows), primarily photometry and astrometry, but also some astrophysical parameters
- XP coefficients, as before, but now with the complete tables including the full covariance matrices (2.6TB, 220M rows)
- RVS spectra, (18GB, 1M rows), R~11,500 spectra


al-jshen:
Test for this dataset passes but fails in the GH action because the runner is out of disk space. Would be nice if https://github.com/MultimodalUniverse/MultimodalUniverse/pull/57 was merged in so that only the test for this dataset runs.

PR #118, opened 2024-07-25
Update Legacy Survey dataset

### LINKED ISSUE #110: Add mask datasets

LTMeyer:
This PR supersedes #115

# Adds Mask and Catalog Information

It adds masks and catalog information to the legacy survey dataset.

Update has already been uploaded to [HG dataset hub](https://huggingface.co/datasets/MultimodalUniverse/legacysurvey). Although the mask object image appears only black on the hub, it actually contains the information as expected. One can check with the following that we correctly obtain the images below:

```python
import datasets
import numpy as np
import matplotlib.pyplot as plt


dataset = datasets.load_dataset("MultimodalUniverse/legacysurvey", split="train")
plt.imshow(np.array(dataset[6]["model_image"]))
plt.show()
plt.imshow(np.array(dataset[6]["object_mask"]))
plt.show()
```
image
image

# Data Generation Fix

On rusty the data generation has been fixed by:
- Checking problematic sweep catalogs;
- Downloading corrupted catalogs.

It is not excluded that some images were missing, which may have resulted in corrupted entries in the dataset.
We ran the `sanity_check.py` script successfully. This doesn't mean however that data are perfectly safe.

Data are available on rusty at the following path `/mnt/ceph/users/polymathic/AstroPile_v1/legacysurvey_update/`

# Other Script Approach

In #115 and the related branch, we experimented with dask. We managed to make it work. However, the parallelism was poor because of locks on concurrent writings on HDF5 files. Consequently, we decided to keep with the current version of the script.

Closes #110


EiffL:
In the meantime, I've written this script to identify and remove files that were corrupted in the previous version of the data:
https://gist.github.com/EiffL/d4ab8e7964572b7453bef81b1d863207

This was necessary for the cross matching script. Overall it removed like 40 files out of 1400.

Interestingly, some files had corrupted values, but did not raise an inflate error, so would go silent in the dataset loading script. We could run the same logic on the new export for a minimal check that everything is ok.

scripts/legacysurvey/build_parent_sample.py
```diff
@@ -139,25 +292,59 @@ def _processing_fn(args):
...
+                    f'{obj["BRICKNAME"]}-{obj["OBJID"]}', dtype=_utf8_filter_typeb
+                ),
+                "gid": obj["gid"],
+                "image_band": np.array(
+                    [f.lower().encode("utf-8") for f in _filters],
+                    dtype=_utf8_filter_type,
+                ),
+                "image_ivar": invvar,
+                "image_array": image,
+                "bit_mask": mask.astype("bool"),
```
EiffL, line 335: it would be better to keep the previous convention of `image_mask`, that's the convention we use in the other image datasets to refer to the "mask" field of an image

EiffL: because the convention for underscore is to mean a subfield of the input field.

EiffL: I went ahead and renamed the field, also renamed the field in all files on ceph, no need to rerun the data preparation script.

EiffL:
I have also renamed `model_image` to `blobmodel` and updated the files on rusty consequently. The rational here is that the pipeline calls this image a blobmodel, it's better to keep the original name.

LTMeyer:
I think we are good for this PR. Should we fix the tests and merge? Regarding the test I don't understand why they are failing.

scripts/legacysurvey/build_parent_sample.py
```diff
@@ -1,39 +1,174 @@
...
+
+def select_observations(catalog, zmag_cut=21.0) -> List[bool]:
+    """Selection function applied to retrieve relevant observation from the DECaLS DR10 South catalog.
+    Observation are deemed relevant based on minimum magnitude, availability in all bands and bit masks.
     """
     # Magnitude cut
-    mask_mag = (22.5 - 2.5*np.log10(catalog['FLUX_Z']/catalog['MW_TRANSMISSION_Z'])) < zmag_cut
+    magnitude = np.zeros_like(catalog["FLUX_Z"])
+    positive_flux_indices = catalog["FLUX_Z"] > 0.0
+    magnitude[positive_flux_indices] = 22.5 - 2.5 * np.log10(
```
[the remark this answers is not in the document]
EiffL: I reverted back to the previous code

EiffL:
I've pushed a few commits doing the following:
- fix the magnitude selection bug that preserved negative z fluxes
- changed the mask code to generate ellipcies instead of disks
- added shape parameters to the catalogs 
- added rgb image as a another "modality" in addition to blob model
- changed the compression filter to `lzf` which the h5py documentation says: `Low to moderate compression, very fast.`

Here is an example of the updated mask and the 2 rgb images:
image

and with the shape information in the catalogs, we can build "bounding boxes" offline:
image


I'm currently regenerating the data using 20 nodes, so should be pretty fast

scripts/legacysurvey/build_parent_sample.py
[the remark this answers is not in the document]
LTMeyer: Good catch, I think I had implemented this because of the complaints on the negative values passed to the log. But I didn't realize it would impact the mask for negative values of the flux.

LTMeyer:
Really nice ! 
At some point I was trying to get the ellipsis instead of the disk for the objects, but I couldn't find a good way to set the ellipsis aspect ratio. I saw that you set it to 2 (c.f. below). Do you think it will fit most of the cases (i.e. there will not be images where the factor 2 is either too small or too big)?

```diff
diff --git a/scripts/legacysurvey/build_parent_sample.py b/scripts/legacysurvey/build_parent_sample.py
index b205fcc..555b0ce 100644
--- a/scripts/legacysurvey/build_parent_sample.py
+++ b/scripts/legacysurvey/build_parent_sample.py
@@ -103,24 +102,22 @@ class CatalogSelector:
         centers = np.stack([j, i]).T
         radii = self.catalog["SHAPE_R"].value / ARCSEC_PER_PIXEL
         object_types = self.catalog["TYPE"].value.astype(str)
-        for c, r, t in zip(centers, radii, object_types):
-            # Enforce a minimum size of the disk mask
-            r = max(2, r)
-            rr, cc = skimage.draw.disk(c, r, shape=self.cutout.shape)
+        e1 = self.catalog["SHAPE_E1"].value
+        e2 = self.catalog["SHAPE_E2"].value
+        angle = 0.5 * np.arctan2(e2, e1)
+        q = (1 - np.sqrt(e1**2 + e2**2)) / (1 + np.sqrt(e1**2 + e2**2))
+        height = 2 * radii
+        width = 2 * radii * q
+        for c, h, w, a, t in zip(centers, height, width, angle, object_types):
+            rr, cc = skimage.draw.ellipse(c[0], c[1], w, h, shape=self.cutout.shape, 
+                                          rotation=a)
```

EiffL (approved):
LGTM

EiffL:
I think it should be fine for most objects. We only keep in the catalog objects that are marked as "extended" meaning they should be larger than point sources. The factor 2 is just the conversion to radius to diameter essentially.

I think it all looks good, no? ready to merge?

PR #122, opened 2024-09-20
Adds Galaxy Zoo 3D Segmentation Data

mb010:
Adding segmentation maps for MaNGA galaxies annotated by humans - collected by [Galaxy Zoo 3D](https://www.zooniverse.org/projects/klmasters/galaxy-zoo-3d) and published in [Masters et al. 2022](https://doi.org/10.1093/mnras/stab2282).  There are a total of 30k samples - most without bar of spiral features.

HuggingFace Dataset Format is outlined in the updated DESIGN.md file.
Each example contains:

The dataset structure is organized with the following features:

Metadata:
- `total_classifications`: The total number of Galaxy Zoo classifications made for this target. Note this is not the number of segementations but the number of citizen scientists who voted on if the galaxy is a spiral or barred galaxy.
- `healpix`: The healpix value in which the subject lies.
- `ra`: Right ascension of the target in degrees.
- `dec`: Declination of the target in degrees.
- `object_id`: The MaNGA ID (unique to each galaxy, not unique to each plate IFU).

Image data:
- `image`: The RGB false color image of the target (MaNGA) galaxy. Data format: `[H,W,C] = [512,512,3]` 16-bit integer.
- `scale`: The angular extent of a pixel in the image in degrees.
- `segmentation`: The segmentation map data. Each key refers to an ordered list of four elements each referring to a specific class.
    - `class`: The class of the segmentation map. One of `center, star, spiral, bar`
    - `vote_fraction`: The fraction of `total_classifications` that were positive towards this class for a class of spiral or bar (e.g. `20/50` total classifications state the subject is barred -> `0.4`). `-1.0` flags that the class is either center or star as there are no votes for these classes. `-0.5` flags if the total_classifications are 0 for the subject.
    - `array`: The segmentation map itself. Values range between 0 and 15, with the value at a given pixel indicating the number of users who included that region in their annotation for the respective class. I.e. a value of 5.0 indicates 5 of the 15 users marked segmentation masks which included that pixel. Data format: `[H,W,C] = [512,512]` 8-bit integer.


scripts/gz3d/test.sh
```diff
@@ -0,0 +1,22 @@
...
+if python build_parent_sample.py --tiny; then
+    echo "Building parent sample for gz3d succeeded"
+else
+    echo "Building parent sample for gz3d failed"
+    exit 1
+fi
+
+# try to load the dataset
+echo "Loading dataset"
+if python -c "from datasets import load_dataset; dset = load_dataset('./gz3d.py', trust_remote_code=True, split='train', streaming=True).with_format('numpy'); next(iter(dset));"; then
```
tom-hehir, line 17: Setting format as numpy probably unnecessary?

tom-hehir (commented):
Looks good! Left a couple of minor comments.

I think the proposed format for segmentation maps is sufficiently general to handle any foreseeable applications to other datasets.

scripts/gz3d/README.md
```diff
@@ -0,0 +1,83 @@
...
+<!-- We suggest downloading the data using globus from a registered host. -->
+
+<!-- ### sdss-access
+
+Alternatively, you can use the `access_tranfer` Python script to download the data with the [sdss-access](https://sdss-access.readthedocs.io/en/latest/) Python package, which uses parallelized `rsync` streams to download the data. This tool organizes the output download directory structure to mirror the official SDSS SAS.
+
+To download all summary files, cubes and maps, run:
+```bash
+python access_transfer.py --destination_path .
+``` -->
```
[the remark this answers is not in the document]
mb010: This is legacy in case I decided to implement it. I tried and it turns out that getting sdss-access to work for this data was not worth it, given the current downloads work just fine (if a little bit slow). I think the solution is the globus from flatiron, but I don't know how to set that up / write about it.

tl;dr: `sdss-access` not actually used. 
```suggestion
```

mb010:
Note: Compare the orientation and scale to legacy survey data.

DESIGN.md
```diff
@@ -147,6 +153,30 @@ We zero-pad multi-band time series data (e.g., for time domain astrophysics) to
...
+        'segmentation': {
+            'class': ['center', 'star', 'spiral', 'bar'],
+            'vote_fraction': [-1.        , -1.        ,  0.75      ,  0.51666667],
+            'array': array(4, 512, 512)
+            },
+        'object_id': '1-100017'
+    }
+  ```
+
+Our main condition is that the data can be mapped back to the relevant survey, and so the segmentation array sizes must match the relevant data from which the maps were derived.
```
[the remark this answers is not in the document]
mb010:
```suggestion
Our main condition for segmentation maps, is that the data can be _mapped back_ to the relevant survey. Specifically, the `segmentation.array` sizes and pixel scale must match the relevant data from which the maps were derived.
```

DESIGN.md
```diff
@@ -147,6 +153,30 @@ We zero-pad multi-band time series data (e.g., for time domain astrophysics) to
...
+        'segmentation': {
+            'class': ['center', 'star', 'spiral', 'bar'],
+            'vote_fraction': [-1.        , -1.        ,  0.75      ,  0.51666667],
+            'array': array(4, 512, 512)
+            },
+        'object_id': '1-100017'
+    }
+  ```
+
+Our main condition for segmentation maps, is that the data can be _mapped back_ to the relevant survey. Specifically, the `segmentation.array` sizes and pixel scale must match the relevant data from which the maps were derived.
```
tom-hehir, line 177:
```suggestion
Our main condition for segmentation maps, is that the data can be _mapped back_ to the relevant survey. Specifically, the `segmentation.array` sizes and pixel scale must match the relevant data from which the maps were derived. Furthermore, the spatial coordinates of the pixels in the segmentation map should implicitly match the spatial coordinates in the relevant survey (which also implies that the segmentation maps and survey have the same orientation).
```

mb010:
```suggestion
Our main condition for segmentation maps, is that the data can be _mapped back_ to the relevant survey. Specifically, the `segmentation.array` sizes and pixel scale must match the relevant data from which the maps were derived. For example, the spatial coordinates of the pixels in the segmentation map should directly match the spatial coordinates of the corresponding pixels in the relevant survey, (which also implies that respective arrays are oriented in the same way).
```

tom-hehir (approved):
LGTM, nice work @mb010

scripts/gz3d/gz3d.py
```diff
@@ -76,17 +76,13 @@ def _info(self):
...
-            'dec': Value('float32'),
+            'total_classifications': Value("uint8"),
             # Image data
-            'image': Array2D(shape=(self._image_size, self._image_size, 3), dtype='int16'),
-            'scale': Value('float32'),
+            'image': Array2D(shape=(self._image_size, self._image_size), dtype='uint8'),
             'segmentation': Sequence(feature={
                 'class': Value('string'),
                 'vote_fraction': Value('float32'),
-                'array': Array2D(shape=(self._image_size, self._image_size), dtype='float32')
```
mb010: I think we should include all of these except healpix (I like the new dtypes more).

mb010 (commented):
Lots of improvements. I do think we need to not drop as many of the features that were originally included in the final data.

EiffL:
[ ] @oliu-io reported that the images are not aligned with the legacysurvey images, apparently there is flip, we should fix that.

scripts/gz3d/build_parent_sample.py
```diff
@@ -0,0 +1,172 @@
...
+        entry.columns = map(str.lower, entry.columns) # lowercase column names
+        entry = Table.from_pandas(entry)
+
+        entry['false_color'] = hdul[0].data.astype(np.uint8)[None, ...]
+        entry['center'] = hdul[1].data.astype(np.uint8)[None, ...]
+        entry['star'] = hdul[2].data.astype(np.uint8)[None, ...]
+        entry['spiral'] = hdul[3].data.astype(np.uint8)[None, ...]
+        entry['bar'] = hdul[4].data.astype(np.uint8)[None, ...]
+    return entry
+
```
tom-hehir, line 105: Add helper function to flip all fields vertically:
```suggestion


def _vertical_flip_fields(entry):
    """Vertically flip fields to be consistent with the orientation of `legacysurvey` fields."""
    entry["false_color"] = np.flipud(entry["false_color"])
    entry["center"] = np.flipud(entry["center"])
    entry["star"] = np.flipud(entry["star"])
    entry["spiral"] = np.flipud(entry["spiral"])
    entry["bar"] = np.flipud(entry["bar"])
    return entry


```

scripts/gz3d/build_parent_sample.py
```diff
@@ -0,0 +1,172 @@
...
+        local_data_path = args.output_dir
+    else:
+        local_data_path = args.local_data_path
+
+    # Read all files in parallel and aggregate the data into a single dataframe
+    locs = list(glob.glob(f"{local_data_path}/*.fits.gz"))
+
+    print(f"Found {len(locs)} FITS files")
+    with Pool() as pool:
+        entries = list(tqdm(pool.imap(_read_fits_file, locs), total=len(locs), desc="Reading FITS files"))
```
tom-hehir, line 131: `gz3d` fields are vertically flipped relative to `legacysurvey` fields. This change would flip the `gz3d` fields to be consistent with the `legacysurvey` fields. Note: this uses the `_vertical_flip_fields` function defined in another suggestion.
```suggestion
        entries = list(tqdm(pool.imap(lambda loc: _vertical_flip_fields(_read_fits_file(loc)), locs), total=len(locs), desc="Reading FITS files"))
        # _vertical_flip_fields gives the fields the same orientation as `legacysurvey` fields
```

tom-hehir (commented):
Two things:
1. Suggested a solution to `gz3d` data being oriented differently to `legacysurvey` data (I found that they were *vertically* flipped relative to each other).
image

2. Noted that `vote_fraction` is not implemented.

scripts/gz3d/gz3d.py
```diff
@@ -0,0 +1,145 @@
...
+                # Preparing an index for fast searching through the catalog
+                sort_index = np.argsort(data["object_id"])
+                sorted_ids = data["object_id"][:][sort_index]
+
+                for k in keys:
+                    # Extract the indices of requested ids in the catalog 
+                    i = sort_index[np.searchsorted(sorted_ids, k)]
+
+                    example = {
+                        'total_classifications': data["gz_total_classifications"][i].astype(np.uint8),
```
tom-hehir, line 134:
```suggestion
                        'total_classifications': data["gz_total_classifications"][i].astype(np.uint8),
                        'pixel_scale': 0.099,  # arcsec / pix
```

PR #126, opened 2024-10-09
Resolve MAST timeouts

jeraud:
Fix an issue where timeouts could occur pre-downloading because of a too large query for MAST when retrieving the data product download links.


EiffL (approved):
LGTM! Thanks a lot @jeraud

PR #127, opened 2024-10-18
Mhuertascompany patch 1

mhuertascompany:
trying to filter out faint galaxies in the JWST dataset. PLEASE CHECK. I have not tried the code.
I just added a few extra lines at the beginning of the cut_stamps_fn function


mb010:
This has been superseded by #128 .

PR #128, opened 2024-10-29
V1/jwst debug

mb010:
Various changes that make the code more readable and importantly have made investigating 'missing' data simpler. Splitting the samples across hdf5 files for each healpix region is key, as any given file is now large enough to open and interrogate. The added complexity is worth not having single files being >25GB with thousands of galaxy cutouts and multiple bands. Additionally #127 is merged into this, meaning there is an optional magnitude cutoff, which we apply to remove extremely faint sources.


mb010:
- [ ] Add mask data

scripts/jwst/build_parent_sample.py
```diff
@@ -32,18 +35,15 @@ def get_pixel_scale(header):
     # Calculate the pixel scale for each direction, assuming square pixels and small angles.
     # We'll convert from degrees to arcseconds by multiplying by 3600.
     pixel_scale_x = np.sqrt(cd[0, 0] ** 2 + cd[1, 0] ** 2) * 3600  # X direction
+    pixel_scale_y = np.sqrt(cd[0, 1] ** 2 + cd[1, 1] ** 2) * 3600  # Y direction
```
LTMeyer, line 38: Can we not group these lines:
```suggestion
    pixel_scale = np.sqrt(cd[0] ** 2 + cd[1] ** 2) * 3600
```

scripts/jwst/build_parent_sample.py
```diff
@@ -171,40 +195,36 @@ def _cut_stamps_fn(
...
-                            idvec.append(idn)
+                        stamp = Cutout2D(sci, position, int(image_size*1.5), wcs=wcs, mode='partial', fill_value=0)
+                        norm = zoom(stamp.data, zoom_factor, order=1)
+                        # Extract central [image_size x image_size] pixels
+                        center = norm.shape[0] // 2
+                        start = center - image_size // 2
+                        end = start + image_size
+                        # Pad if necessary
+                        if np.max(norm) <= 0:
+                            # print(f"Object {idn:8d} has invalid stamp. Appending blank image. {np.count_nonzero(norm == 0)}, {norm.shape}")
```
LTMeyer, line 212: Can we remove commented prints?

scripts/jwst/build_parent_sample.py
```diff
@@ -267,172 +288,125 @@ def _processing_fn(args):
...
+    print("... Finished processing all groups")
+    return 1
 
-        # Join on object_id with the input catalog
-        catalog = join(catalog, images, keys="object_id", join_type="inner")
 
-        # Making sure we didn't lose anyone
-        assert len(catalog) == len(images), "There was an error in the join operation"
+# Initial survey information
+from info import _SURVEYS_INFO
```
LTMeyer, line 375: Why is this import in the middle of the file?

LTMeyer (commented):
Few shallow comments. I didn't check the whole data prcoessing logic as it seems it didn't change much.

scripts/jwst/build_parent_sample.py
mb010: It's not - I was just being a lazy engineer.

LTMeyer (approved):

PR #129, opened 2024-10-29
V1/manga debug

mb010:
Enabling the chunking of samples into multiple hdf5 files per healpix region to prevent (presumptive) memory errors. The largest single healpix region is ~326GB post processed (healpix=219). Given that this data was being formatted in parallel with up to 10 workers, even the relatively big nodes (with 1000GB of memory) likely had memory overflow errors. I did not encounter these with the file chunking and reducing the number of workers.

This replaces #109.


mb010:
The current version produces an error from the test:
```shell
Traceback (most recent call last):
  File "<string>", line 1, in <module>
StopIteration
Loading manga dataset failed
```
On a local version, the tests pass. (although an initial test failed due to a mismatch between available C++ package and torch version - downgrading torch solved this for me).

I'm open to suggestions on how to fix this.

havok2063:
@mb010 Is the error you're seeing something specific to the manga dataset?  if so, can you tell which target is failing?  the traceback looks pretty general.

mb010:
The error comes straight from the failed checks of the github actions. It doesn't give any more context and I cannot reproduce it locally - my local tests complete with no issue.
Maybe @LTMeyer has an idea of how to look into this?

PR #138, opened 2024-11-11
Brian/tess_pipelines - additional pipelines for TESS data.

Birr0:
This PR includes other pipelines to produce light curves from TESS full frame images. All of these pipelines are available on MAST. An overview of the features included in this request are:

* QLP, TGLC and SPOC (already available on MMU) pipelines have been added.

* Each pipeline inherits from a general, parent class which can download data from MAST; it is hoped this class to make it easier to download from additional pipelines or other sources in the future with minimal additional code required.

* Async requests have been implemented along with distributed processing of data to help improve speed and efficiency.

* Tests have also been written to ensure the expected data output is achieved.


mb010 (changes requested):
@Birr0 can you restructure so that your build_parent_sample.py and spoc.py are in the scripts/tess folder (and rename as necessary).

@lhparker1 @Birr0 mentioned that the [base_downloader](scripts/tess/pipelines/base_downloader.py) script might be a good utils for parallel MAST downloading in general? Can/should we put it in the [mmu utils](mmu/utils.py) for future use?

Birr0:
> @Birr0 can you restructure so that your build_parent_sample.py and spoc.py are in the scripts/tess folder (and rename as necessary).
> 
> @lhparker1 @Birr0 mentioned that the [base_downloader](scripts/tess/pipelines/base_downloader.py) script might be a good utils for parallel MAST downloading in general? Can/should we put it in the [mmu utils](mmu/utils.py) for future use?

@mb010 The file structure should now be amended for this purpose.

scripts/tess/README.md
```diff
@@ -1,35 +1,101 @@
...
+To download a full sector and store it in the standardised format, simply call:
 
-In the current version of the dataset, we retrieve all optical spectra from the TESS SPOC (Science Processing Operations Center) pipeline for sector 64. There are currently no specific cuts applied, but these could be specified in the mast_s3_transfer.py file.
+```
+python build_parent_sample.py --pipeline ['qlp', 'tglc', 'spoc'] -s [sector_number] --data_path './data' 
+--hdf5_output_path './data/' 
+--fits_output_path './data/fits' --n_processes 4 --tiny
+```
+
+The ```-s``` flag determines which sector to download (**i.e. sector 23**). The following flags: ```--data_path```, ``` --hdf5_output_path``` and ```--fits_output_path ``` each specify where the root data path, where the standardised hdf5 files are saved and where the fits files from MAST are kept. ```--n_processes``` allows the downloads and processing of the data to be distributed across cores. ```--tiny``` is a boolean flag which can be used for testing, which currently uses 100 samples. This can be changed by modifying ```_TINY_SIZE``` in ```build_parent_sample.py```.
```
lhparker1, line 15: Can you specify which sectors are available?

scripts/tess/README.md
```diff
@@ -1,35 +1,101 @@
-# TESS Dataset Collection
 
-This folder contains the scripts and queries used to build the TESS photometric parent sample, based on 
-a set of selection criteria for the TESS-SPOC light curves.
+# Dataset builders for the TESS FFI pipelines
 
-## Sample selection
+This folder contains scripts used to build lightcurve datasets from a number of pipelines including: TGLC developed by [Han & Brandt 2023](https://iopscience.iop.org/article/10.3847/1538-3881/acaaa7), QLP ([Huang et al. 2020](https://arxiv.org/abs/2011.06459)) and SPOC ([Caldwell et al. 2020](https://ui.adsabs.harvard.edu/abs/2020RNAAS...4..201C/abstract)). Note, currently there are speed-ups in download times with distributed downloading implemented in this version.
```
lhparker1, line 4: It would be helpful to specify at a high level what the difference between these pipelines is (if I have never worked with this data before, which one might I want to use for which application).

scripts/tess/spoc.py
```diff
@@ -0,0 +1,255 @@
...
+    def _info(self):
+        features = {"lightcurve" : Sequence({
+            'time': Value(dtype="float32"),
+            'flux': Value(dtype="float32"),
+            'flux_err':  Value(dtype="float32"),
+            'quality':  Value(dtype="float32")
+            }),
+            'RA':  Value(dtype="float32"),
+            'DEC':  Value(dtype="float32"),
+            'TIC_ID': Value(dtype="string")
```
lhparker1, line 83: this should be `object_id` right?

scripts/tess/qlp.py
```diff
@@ -0,0 +1,309 @@
...
+            'sap_x':  Value(dtype="float32"),
+            'sap_y':  Value(dtype="float32"),
+            'sap_bkg':  Value(dtype="float32"),
+            'sap_bkg_err':  Value(dtype="float32"),
+            'kspsap_flux_sml':  Value(dtype="float32"),
+            'kspsap_flux_lag':  Value(dtype="float32"),
+            }),
+            'RA':  Value(dtype="float32"),
+            'DEC':  Value(dtype="float32"),
+            'TIC_ID': Value(dtype="string"),
```
lhparker1, line 94: this should be `object_id`, right?

lhparker1 (changes requested):
Overall, this looks really good, thank you Brian!

I have one main comment - in keeping with the rest of the MMU, I think it would make more sense to reorganize the 3 huggingface datasets (spoc, tlgc, and qlp) into one (tess), which has different configs for each of these three. Then, we would have one tess.py file that can handle each of these three configs. Also, the downloaders can probably go in their own scripts, as they aren't needed at the huggingface level it seems (and it could be useful to do this config versioning for those too).

@mb010 LMK if that makes sense to you too.

scripts/tess/README.md
Birr0: Updated the README file to the following:

```
The ```-s``` flag determines which sector to download (**i.e. sector 23**). The following sectors are available for the current pipelines (as of 10/02/2025; see above links for updates):
- TGLC: 1-52
- SPOC: 1-83
- QLP: 1-83


The following flags: ```--data_path```, ``` --hdf5_output_path``` and ```--fits_output_path ``` each specify where the root data path, where the standardised hdf5 files are saved and where the fits files from MAST are kept. ```--n_processes``` allows the downloads and processing of the data to be distributed across cores. ```--tiny``` is a boolean flag which can be used for testing, which currently uses 100 samples. This can be changed by modifying ```_TINY_SIZE``` in ```build_parent_sample.py```.

```

scripts/tess/README.md
Birr0: I've added some of the following high-level differences of the pipelines to the docs.
```
The methods used to extract light curves in each pipeline vary signficantly. For detailed reference of their implementations see the linked papers above. At a high level, the SPOC pipeline include targets within 100 pc that are brighter than $H <10$. These targets have been traditionally used for exoplanet searches with de-trended light curves available for this purpose; pre-search data conditioning (PDC) is used to remove long term trends using co-trending basis vectors. Simple aperture photometry is also available and is recommended for applications focused on stellar variability.

The QLP pipeline focuses on similar data products as SPOC providing simple aperture photometry measurements. Detrending of stellar lightcurves using a spline-based method is also produced in the data products offering a different methodology to SPOC.

The TGLC pipeline uses GAIA DR3 as priors to extract targets from TESS full-frame images that are on the order of 2.5 (TESS) magntiudes dimmer than those from SPOC/QLP and demonstrates enhanced deblending capabilities. These increases in processing have increased the number of availble lightcurves from TESS by at least an order of magnitude.
```

scripts/tess/qlp.py
Birr0: Changed this to object_id to match MMU specifications.

scripts/tess/README.md
lhparker1: Looks good, thanks!

scripts/tess/README.md
lhparker1: LGTM!

mb010:
I think we have addressed most things @lhparker1 . The biggest open question is renaming to default keys for lightcurves for the non-default pipelines (QLP and TGLC). I.e. QLP and TGLC pipelines currently dont have `flux` and `flux_err` keys.

mb010 (approved):

PR #116, opened 2024-07-08
Citation script

mariel-pettee:
Added a basic script containing acknowledgement & citation info for each dataset. I did my best to be as thorough as possible, but it's possible there are some mistakes here. I think it's a solid first draft, though!

PR #140, opened 2024-11-18
Information Retrieval: Making Citing and Acknowledgements Easy

mb010:
Based on the original ideas in #116, this PR introduces `scripts/info.py` as a utility to retrieve, save, and inspect all of the citations and acknowledgements for the underlying data.


mb010:
License field currently unfilled for:
  - [ ] chandra
  - [ ] gaia
  - [ ] jwst
  - [ ] legacysurvey
  - [ ] sdss
  - [ ] vipers
 
 cfa and new PRs need to be checked against info.py to make sure their citations and acknowledgements are stored correctly.

DESIGN.md
```diff
@@ -162,18 +162,23 @@ import h5py
...
-title = {A great new dataset},
-author={huggingface, Inc.
-},
-year={2020}
+    title = {A great new dataset},
+    author={huggingface, Inc.
+    },
+    year={2020}
 }
 """
```
LTMeyer, line 175: Is the citation supposed to point to a HF publication or to the original paper?

README.md
```diff
@@ -94,6 +94,32 @@ Illustration of the methodology behind the Multimodal Universe. Domain scientist
...
+
+It allows you to retrieve all of the dataset information, or just acknowledgement and citation information for some or all datasets. By not specifying a dataset, it will return all datasets. By not specifying at least one of `--citation` or `--acknowledge`, it will return all of the information (including license, homepage, etc.).
+```sh
+python scripts/info.py --cite --data <datasets>
+python scripts/info.py --acknowledge --data <datasets>
+```
+For example, to get the citations for the APOGEE and SDSS datasets and save them to `info_citation.txt`, run:
+```sh
+python scripts/info.py --cite --data apogee sdss -o info_citation.txt
+```
```
LTMeyer, line 109: Would it be preferable to have the output as a `.bib`? If it's this format anyways we may mention it in the example.

```suggestion
For example, to get the citations for the APOGEE and SDSS datasets and save them to `info_citation.bib`, run:
```sh
python scripts/info.py --cite --data apogee sdss -o info_citation.bib
```
```

File: scripts/info.py [hunk 3]
```diff
@@ -0,0 +1,118 @@
...
+def get_all_datasets() -> List[str]:
+    # Collect all unique dataset names from the scripts directory
+    all_datasets = set()
+    # only consider directories in the scripts directory
+    for file in os.listdir('scripts'):
+        if os.path.isdir(os.path.join('scripts', file)):
+            all_datasets.add(file)
+    return sorted(list(all_datasets))
+
+def get_info(dataset: str, info_keys: List[str] = ['citation']) -> dict:
```
LTMeyer, line 19: What is the purpose of this function? Can we comment it in a docstring?

scripts/info.py
```diff
@@ -0,0 +1,118 @@
...
+                    dataset_info["acknowledgements"] = "\n".join([line.lstrip("% ") for line in info.citation.split("% CITATION\n")[0].split("\n")])
+                else:
+                    dataset_info["acknowledgements"] = "% ACKNOWLEDGEMENTS\nNot available"
+        for key in info_keys:
+            if key == "acknowledgements":
+                continue
+            value = getattr(info, key, "Not available")
+            dataset_info[key] = value if value else "Not available"
+        return dataset_info
+    except Exception as e:
```
LTMeyer, line 50: Could we avoid long try and except clauses? What is the problem that we may face and we want to catch?

scripts/info.py
LTMeyer, line 19: Can we not retrieve these information from the instantiated dataset directly?

LTMeyer (commented):
Few questions to know whether we could simplify the script.

DESIGN.md
mb010: To the paper that should be cited if the dataset is used.

README.md
mb010: Yeah. The example should probably do that. (Encoding wise they are the same afaik).

scripts/info.py
mb010: I have refactored to make the process clearer and only try/except the data loading which is what I was initially worried about. Thanks for catching this.

scripts/info.py
mb010: Updated doc strings.

LTMeyer (approved):

PR #141, opened 2024-11-20
apogee updates

al-jshen:
fix how mask is created, fix inconsistencies in some column names

mask now created from `hdus[3].data[0]` (int bitmask) of spectra fits files instead of `hdus[2].data[0]` (error/variance array) according to the [data model](https://data.sdss.org/datamodel/files/APOGEE_REDUX/APRED_VERS/stars/TELESCOPE/FIELD/apStar.html) 

for consistency with other spectral datasets, spectral columns are now all `spectrum_XXX` and the mask is called `mask` instead of `pix_bitmask`


mb010 (approved):
LGTM

PR #147, opened 2024-12-02
Update of JWST to simplify script, enforce all color, and improve inverse variance mask

EiffL:
This PR adds the following elements:
- it improves the computation of the inverse variance weights for JWST following the instruction here https://dawn-cph.github.io/dja/blog/2023/07/18/image-data-products/ 
- it removes the resampling of images to preserve native pixel scale (MMU should not modify the data from the original source, at most we apply cuts).
- Applies full color cut so that we don't have objects with missing bands
- Streamline the downloading and processing.


scripts/jwst/jwst.py
```diff
@@ -361,20 +288,18 @@ def _generate_examples(self, files, object_ids=None):
...
-                                    ~np.isfinite(data["image_array"][i][j])
-                                ).astype(bool),
+                                "flux": data["image_flux"][i][j],
+                                "ivar": data["image_ivar"][i][j],
+                                "mask": data["image_mask"][i][j].astype(bool),
                                 "psf_fwhm": data["image_psf_fwhm"][i][j],
                                 "scale": data["image_scale"][i][j],
                             }
-                            for j, _ in enumerate(self.config.bands)
+                            for j in range(len(data["image_band"][i]))
```
mb010, line 297: This should probably throw a warning if the band in the data is not the same as the bands expected by the loader.

scripts/jwst/build_parent_sample.py
```diff
@@ -1,404 +1,352 @@
...
+_mosaics = [
+    # Primer fields
+    "primer-cosmos-east-grizli-v7.0", "primer-cosmos-west-grizli-v7.0",
+    "primer-uds-north-grizli-v7.2", "primer-uds-south-grizli-v7.2",
+    # CEERS fields
+    "ceers-full-grizli-v7.4",
+    # NGDEEP fields
+    "ngdeep-grizli-v7.2",
+    # GOODS fields
+    "gds-grizli-v7.2", "gdn-grizli-v7.3",
```
mb010, line 31: Why aren't we processing / formatting more from the [DJA index](https://s3.amazonaws.com/grizli-v2/JwstMosaics/v7/index.html)?

scripts/jwst/build_parent_sample.py
```diff
@@ -409,35 +357,22 @@ def main(output_dir: str, image_dir: str, survey: str, subsample: str, mag_cut:
...
-    )
-    parser.add_argument(
-        "--mag_cut",
-        type=float,
-        default=25.5,
-        help="Magnitude cut for the sample.",
+        "--max_workers",
+        type=int,
+        default=10,
+        help="Number of threads for parallel downloads and processing",
```
mb010, line 369: Can we default to none and use `os.cpu_count()` as a default?

mb010 (changes requested):
See comments.

scripts/jwst/build_parent_sample.py
EiffL: because these are the fields we documented in the paper

scripts/jwst/build_parent_sample.py
EiffL: os.cpu_count() is too much usually

EiffL: happy to switch default to 1 though
```suggestion
        "--max_workers",
        type=int,
        default=1,
        help="Number of threads for parallel downloads and processing",
```

scripts/jwst/build_parent_sample.py
mb010: Agreed. I think for a v2, it would be good to add support for them. :)

EiffL:
Ok, I confirm it works, and the images it loads are not empty

scripts/jwst/README.md
```diff
@@ -4,12 +4,38 @@ The data is donwloaded from the [DAWN JWST archive](https://dawn-cph.github.io/d
...
+- f115w
+- f150w
+- f200w
+- f277w
+- f356w
+- f444w
+
+The current default image size is 96 pixels. For most surveys the pixel scale is 0.02 arcsec/pixel for the short wavelength bands of NIRCam, and 0.04 arcsec/pixel for the long wavelength bands. 
+
+More information about the source imaging data canb be found here: https://dawn-cph.github.io/dja/blog/2023/07/18/image-data-products/.
```
mb010, line 24:
```suggestion
More information about the source imaging data can be found here: https://dawn-cph.github.io/dja/blog/2023/07/18/image-data-products/. 
```

scripts/jwst/build_parent_sample.py
mb010: Yeah. Let's change the default to 1.

mb010 (approved):
LGTM

PR #148, opened 2024-12-02
Changing name from astropile to mmu

mb010:
Changing name as discussed from `astropile` to `mmu`.


EiffL:
It looks very reasonable to me, but impossible to tell if nothing breaks ^^

EiffL:
Ok, I found one issue in the demo notebook, fixing it

EiffL:
Ok, I've pushed revised notebook

EiffL (approved):
LGTM

PR #149, opened 2024-12-02
Modifications needed for HF sync 

EiffL:
This PR implements a number of consistency fixes to enable the automated upload to HuggingFace. 

These modifications have been tested, and the HF previews are being updated.

#### No renaming/reformatting of native fields

Apart from ra, dec, object_id, image, spectrum, lightcurve, **we do not rename variables from the original surveys**, not even to normalize redshift between Z, z, 'redshift'. 

This is important because **we do not provide documentation for what these fields mean**, and instead direct the user to the original survey documentation. There is no such thing as a "redshift" for instance, any redshift measurement is the output of some algorithm, same for magnitude and such. It is important that the user go back to the survey-specific definition, otherwise we need to do a lot more work to add these descriptions ourselves.

#### Reversionning all datasets to 1.0.0

When  we make updates after this, we can incrementally increase this version number

### DIFF
--- a/scripts/btsbot/btsbot.py
+++ b/scripts/btsbot/btsbot.py
@@ -68,13 +68,11 @@ _HOMEPAGE = "https://zenodo.org/records/10839691"
 
 _LICENSE = "CC BY 4.0"
 
-_VERSION = "10.0.0"
+_VERSION = "1.0.0"
 
 _FLOAT_FEATURES = [
     'jd',
     'diffmaglim',
-    'ra',
-    'dec',
     'magpsf',
     'sigmapsf',
     'chipsf',
@@ -154,7 +152,6 @@ _INT_FEATURES = [
     'nmtchps',
     'nnotdet',
     'N',
-    'healpix',
 ]
 
 _BOOL_FEATURES = [
@@ -180,9 +177,9 @@ class BTSbot(datasets.GeneratorBasedBuilder):
             name="BTSbot", 
             version=VERSION,
             data_files=DataFilesPatternsDict.from_patterns({
-                'train': ['./data/healpix=*/train_001-of-001.hdf5'],
-                'val': ['./data/healpix=*/val_001-of-001.hdf5'],
-                'test': ['./data/healpix=*/test_001-of-001.hdf5'],
+                'train': ['data/healpix=*/train_001-of-001.hdf5'],
+                'val': ['data/healpix=*/val_001-of-001.hdf5'],
+                'test': ['data/healpix=*/test_001-of-001.hdf5'],
                 }),
             description="BTSbot dataset with train, val, and test splits."
             ),
@@ -268,10 +265,10 @@ class BTSbot(datasets.GeneratorBasedBuilder):
                     example = {
                         'image': [
                             {
-                                'band': data['band'],
+                                'band': data['band'][i],
                                 'view': view,
                                 'array': data['image_triplet'][i, :, :, j],
-                                'scale': data['image_scale'],
+                                'scale': data['image_scale'][i],
                             }
                             for j, view in enumerate(self._views)
                         ]
--- a/scripts/cfa/cfa.py
+++ b/scripts/cfa/cfa.py
@@ -18,14 +18,14 @@ import datasets
 from datasets import Features, Value, Sequence
 from datasets.data_files import DataFilesPatternsDict
 
-_VERSION = "0.0.1"
+_VERSION = "1.0.0"
 
 _HOMEPAGE = "https://lweb.cfa.harvard.edu/supernova/"
 _LICENSE = "CC BY 4.0"
 
 # Features common across all CFA datasets
 _STR_FEATURES = ["object_id", "obj_type"]
-_FLOAT_FEATURES = ["ra", "dec"]
+_FLOAT_FEATURES = []
 
 # Dataset-specific descriptions and citations
 _DESCRIPTIONS = {
--- a/scripts/gz10/gz10.py
+++ b/scripts/gz10/gz10.py
@@ -62,7 +62,7 @@ _VERSION = "1.0.0"
 class GZ10(datasets.GeneratorBasedBuilder):
     """GZ-10 Catalog loaded with Healpix indices (using NSIDE=16). uint8 images included."""
 
-    VERSION = datasets.Version("0.0.1")
+    VERSION = _VERSION
 
     BUILDER_CONFIGS = [
         datasets.BuilderConfig(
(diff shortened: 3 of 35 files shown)


scripts/btsbot/btsbot.py
```diff
@@ -268,10 +265,10 @@ def _generate_examples(self, files, object_ids=None):
                     example = {
                         'image': [
                             {
-                                'band': data['band'],
+                                'band': data['band'][i],
                                 'view': view,
                                 'array': data['image_triplet'][i, :, :, j],
-                                'scale': data['image_scale'],
+                                'scale': data['image_scale'][i],
```
mb010, line 271: Has this been tested?

scripts/btsbot/btsbot.py
```diff
@@ -68,13 +68,11 @@
 
 _LICENSE = "CC BY 4.0"
 
-_VERSION = "10.0.0"
+_VERSION = "1.0.0"
```
mb010, line 71: Note: It was v10 to match the official release of BTSbot data.

scripts/cfa/cfa.py
```diff
@@ -18,14 +18,14 @@
...
-_VERSION = "0.0.1"
+_VERSION = "1.0.0"
 
 _HOMEPAGE = "https://lweb.cfa.harvard.edu/supernova/"
 _LICENSE = "CC BY 4.0"
 
 # Features common across all CFA datasets
 _STR_FEATURES = ["object_id", "obj_type"]
-_FLOAT_FEATURES = ["ra", "dec"]
+_FLOAT_FEATURES = []
```
mb010, line 28: Are these not used?

scripts/gz10/gz10.py
```diff
@@ -93,9 +93,7 @@ def _info(self):
         features = datasets.Features(
             {
                 "gz10_label": datasets.Value("int32"),
-                "ra": datasets.Value("float32"),
-                "dec": datasets.Value("float32"),
```
mb010, line 97: Why are we dropping these keys? I think they should be included for cross matching.

scripts/gz10/gz10.py
```diff
@@ -150,9 +148,7 @@ def _generate_examples(self, files, object_ids=None):
 
                     example = {
                         "gz10_label": data["ans"][i].astype(np.int32),
-                        "dec": data["dec"][i].astype(np.float32),
-                        "ra": data["ra"][i].astype(np.float32),
```
mb010, line 154: Why are we dropping these keys? I think they should be included for cross matching.

scripts/manga/manga.py
```diff
@@ -81,9 +81,6 @@ def _info(cls):
 
         # add metadata features
         features['object_id'] = Value("string")
-        features['ra'] = Value("float32")
-        features['dec'] = Value("float32")
```
mb010, line 85: These shouldn't be dropped unless I'm missing something.

scripts/snls/snls.py
```diff
@@ -61,16 +61,14 @@
...
+_VERSION = "1.0.0"
 
 _STR_FEATURES = [
     "object_id",
     "obj_type"
 ]
 
 _FLOAT_FEATURES = [
-    "ra", 
-    "dec",
```
mb010, line 73: These shouldn't be dropped unless I'm missing something.

scripts/ps1_sne_ia/ps1_sne_ia.py
```diff
@@ -88,7 +86,7 @@ class PS1SNIa(datasets.GeneratorBasedBuilder):
         datasets.BuilderConfig(
             name="ps1_sne_ia",
             version=VERSION,
-        data_files=DataFilesPatternsDict.from_patterns({"train": ["./healpix=*/*.hdf5"]}), # This seems fairly inflexible. Probably a massive failure point.
+        data_files=DataFilesPatternsDict.from_patterns({"train": ["ps1_sne_ia/healpix=*/*.hdf5"]}), # This seems fairly inflexible. Probably a massive failure point.
```
mb010, line 89:
```suggestion
        data_files=DataFilesPatternsDict.from_patterns({"train": ["ps1_sne_ia/healpix=*/*.hdf5"]}),
```
Removing comment.

scripts/ssl_legacysurvey/ssl_legacysurvey.py
```diff
@@ -92,7 +92,7 @@ class SSLLegacySurvey(datasets.GeneratorBasedBuilder):
             name="stein_et_al",
             version=VERSION,
             data_files=DataFilesPatternsDict.from_patterns(
-                {"train": ["*/healpix=*/*.hdf5"]}
+                {"train": ["north/healpix=*/*.hdf5"]}
```
mb010, line 95: not `ssl_legacysurvey`?

scripts/swift_sne_ia/swift_sne_ia.py
```diff
@@ -60,16 +60,14 @@
...
+_VERSION = "1.0.0"
 
 _STR_FEATURES = [
     "object_id",
     "obj_type"
 ]
 
 _FLOAT_FEATURES = [
-    "ra", 
-    "dec",
```
mb010, line 72: These shouldn't be dropped unless I'm missing something.

mb010 (changes requested):
See comments.

There is a repeating question: Why is this PR trying to drop `ra` and `dec` in many of the datasets?

scripts/gz10/gz10.py
EiffL: no, cross matching is not done at that level

scripts/snls/snls.py
EiffL: we are not including ra, dec as features

scripts/ssl_legacysurvey/ssl_legacysurvey.py
EiffL: no, it's called north

scripts/btsbot/btsbot.py
EiffL: yeah, but this refers to the MMU version of the dataset

scripts/btsbot/btsbot.py
EiffL: yes

scripts/ssl_legacysurvey/ssl_legacysurvey.py
EiffL: because in principle we could have also added a 'south'

mb010 (approved):
LGTM

PR #154, opened 2024-12-11
update galah prep scripts

al-jshen:


scripts/galah/build_parent_sample.py
```diff
@@ -1,391 +1,589 @@
...
+                spectra[i][k],
+                (0, max_length - len(spectra[i][k])),
+                mode="constant",
+                constant_values=-1,
+            )
+
+    spectra = collate(spectra)
+
+    os.makedirs(output_dir, exist_ok=True)
+    with h5py.File(os.path.join(output_dir, "001-of-001.hdf5"), "a") as f:
```
mb010, line 450: I think we should be storing these as chunks in each hdf5 file. I.e. limited to N samples per HDF5 file and actually use the indicies? Even if for some healpix folders this chagnes nothing, it likely will reduce errors later.

scripts/galah/download.py
mb010, line 1: Using beautiful soup might make this more readable and stable. If you want to change it you can, I don't see this as blocking, but its worth thinking about.

scripts/galah/galah.py
```diff
@@ -54,62 +55,280 @@
...
             data_files=DataFilesPatternsDict.from_patterns(
-                {"train": ["galah_dr3/healpix=*/*.hdf5"]}
+                {"train": ["dr3/healpix=*/*.hdf5"]}
             ),
             description="GALAH DR3",
         ),
     ]
 
-    DEFAULT_CONFIG_NAME = "galah_dr3"
+    DEFAULT_CONFIG_NAME = "dr3"
```
mb010, line 313: Should the name be `dr3` or `galah`?

scripts/galah/test.sh
```diff
@@ -1,17 +1,13 @@
 #!/bin/bash
 
-# First build the parent sample and save both raw and H5 to current directory
-if python build_parent_sample.py "./downloaded_galah" "." --num_processes 1 --tiny; then
-    echo "Build parent sample for GALAH successful"
-else
-    echo "Build parent sample for GALAH failed"
-    exit 1
-fi
+python3 download.py --tiny
```
mb010, line 3: I think we have been using `python` throughout with the assumption that people would be working with an appropriate env.
```suggestion
python download.py --tiny
```

scripts/galah/download.py
```diff
@@ -0,0 +1,37 @@
...
+)
+args = parser.parse_args()
+
+os.makedirs(args.output_dir, exist_ok=True)
+
+print("DOWNLOADING VAC, CATALOG, RESOLUTION MAPS, MISSING SPECTRA LIST")
+
+os.system(
+    f'aria2c -j16 -s16 -x16 -c -Z "https://cloud.datacentral.org.au/teamdata/GALAH/public/GALAH_DR3/GALAH_DR3_VAC_ages_v2.fits" "https://cloud.datacentral.org.au/teamdata/GALAH/public/GALAH_DR3/GALAH_DR3_main_allstar_v2.fits" "https://github.com/svenbuder/GALAH_DR3/raw/refs/heads/master/analysis/resolution_maps/ccd1_piv.fits" "https://github.com/svenbuder/GALAH_DR3/raw/refs/heads/master/analysis/resolution_maps/ccd2_piv.fits" "https://github.com/svenbuder/GALAH_DR3/raw/refs/heads/master/analysis/resolution_maps/ccd3_piv.fits" "https://github.com/svenbuder/GALAH_DR3/raw/refs/heads/master/analysis/resolution_maps/ccd4_piv.fits" "https://cloud.datacentral.org.au/teamdata/GALAH/public/GALAH_DR3/spectra/GALAH_DR3_list_missing_reduced_spectra_v2.csv" --dir={args.output_dir}'
+)
```
mb010, line 21: Can we split these or comment them to highlight which is which for future reference? For some its obvious for others its not.

mb010 (changes requested):
Some questions and suggestions.

scripts/galah/download.py
al-jshen: I'll comment them. I originally had them split but I think it's actually a bit faster as written because then aria2 can download them concurrently.

scripts/galah/download.py
al-jshen: Done. Definitely a better option than my really janky bash script lol

scripts/galah/build_parent_sample.py
EiffL: Yes we have this mechanism if we ever run into files that start to be way too big. But for now the cross matching logic assumes there is a single hdf5 file per heal pix region. So for Galah I think we can just have single files, they are probably not that big

al-jshen: Okay, I have implemented the splitting/chunking with a default of a one chunk per healpix. So it should work with the current cross matching logic but we can also update this in the future.

al-jshen: Also yeah for GALAH it doesn't seem to matter too much, for a version I've already built the healpix with the largest number of objects has 2335 rows.

scripts/galah/test.sh
```diff
@@ -1,17 +1,13 @@
...
-# Try to load the dataset with hugging face dataset
-if python -c "from datasets import load_dataset; dset = load_dataset('./galah.py', trust_remote_code=True, split='train').with_format('numpy'); print(f'loaded dataset with {len(dset)} examples'); next(iter(dset));"; then
-    echo "Load dataset for GALAH successful"
-else
-    echo "Load dataset for GALAH failed"
-    exit 1
-fi
+cd data/spectra
+
+../../prepare.sh
```
mb010, line 7:
```suggestion
if ../../prepare.sh; then
    echo "Data preparation (prepare.sh) for GALAH successful"
else
    echo "Data preparation (prepare.sh) for GALAH failed"
    exit 1
fi
```

scripts/galah/test.sh
```diff
@@ -1,17 +1,13 @@
...
+
+cd ../..
+
+python build_parent_sample.py --data_dir data/spectra/spectra --allstar_file data/GALAH_DR3_main_allstar_v2.fits --resolution_map_dir data --vac_file data/GALAH_DR3_VAC_ages_v2.fits  --missing_id_file data/GALAH_DR3_list_missing_reduced_spectra_v2.csv --output_dir ./dr3 --nside 16 --num_workers 4 --tiny
```
mb010, line 11:
```suggestion
if python build_parent_sample.py --data_dir data/spectra/spectra --allstar_file data/GALAH_DR3_main_allstar_v2.fits --resolution_map_dir data --vac_file data/GALAH_DR3_VAC_ages_v2.fits  --missing_id_file data/GALAH_DR3_list_missing_reduced_spectra_v2.csv --output_dir ./dr3 --nside 16 --num_workers 4 --tiny; then
    echo "Build parent sample for GALAH successful"
else
    echo "Build parent sample for GALAH failed"
    exit 1
fi
```

mb010 (approved):
Logging in test time needs to be added back in (suggestions can be committed if you're happy with them) and then we can merge.

PR #156, opened 2024-12-22
Kepler dataset

IlayMalinyak:
Hi all,
I created a Kepler light curves dataset. 
I used the TESS directory as a reference and modified it according to the specifications of the Kepler mission.
One of the differences is that Kepler has two types of flux pipelines - PDCSAP and SAP. while PDCSAP is usually the go-to,  some people might use the raw flux (SAP). Therefore, I used both in the dataset and instead of the regular 'flux' field inside 'lightcurve', I created a 'pdcsap_flux' and 'sap_flux' with the corresponding errors.
Please let me know what you think.

Best,
Ilay


scripts/kepler/run.sh
```diff
@@ -0,0 +1,18 @@
+#SBATCH --job-name=Multimodal-kepler  # create a short name for your job
+#SBATCH --nodes=1                # node count
+#SBATCH --ntasks-per-node=1      # total number of tasks per node
+#SBATCH --cpus-per-task=96        # cpu-cores per task (>1 if multi-threaded tasks)
+#SBATCH --gpus-per-node=1     # number of gpus per node          
+#SBATCH --mail-type=begin        # send email when job begins
+#SBATCH --mail-type=end          # send email when job ends
+#SBATCH --mail-user=ilay.kamai@campus.technion.ac.il
```
helenqu, line 8: do we want to accept this with so much individualized code?

scripts/kepler/download_data.py
```diff
@@ -0,0 +1,70 @@
...
+    if args.tiny:
+        indices = indices[:10]
+    # Get the number of CPUs from SLURM environment variable
+    num_cpus = args.num_processes
+    print(num_cpus, indices)
+    
+    with Pool(num_cpus) as pool:
+        pool.map(process_sample, indices)
+
+def download_data():
```
helenqu, line 57: this method doesn't appear to be used

scripts/kepler/download_data.py
```diff
@@ -0,0 +1,70 @@
...
+    filename='download.log',
+    level=logging.INFO,
+    format='%(asctime)s - %(levelname)s - %(message)s'
+)
+
+all_files = os.listdir('fits_data')
+
+def process_sample(idx):
+    try:
+        all_samples = pd.read_csv('all_kepler_samples.csv')
```
helenqu, line 21: where does this file come from? should this path be passed in too?

scripts/kepler/download_data.py
```diff
@@ -0,0 +1,70 @@
...
+import argparse
+
+# Configure logging
+logging.basicConfig(
+    filename='download.log',
+    level=logging.INFO,
+    format='%(asctime)s - %(levelname)s - %(message)s'
+)
+
+all_files = os.listdir('fits_data')
```
helenqu, line 17: pass this in rather than hardcode

scripts/kepler/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Data preparation
+
+### Downloading data through MAST/Astroquery
+
+The first step to data preparation is to download all relevant data to a local machine through MAST/lightkurve. By default it will currently download the long cadence data.
+
+You can download the data through the following script
+```bash
+python download_data.py [--tiny] [path to catalog with KICs]
```
helenqu, line 18: how do we obtain this "catalog with KICs"?

scripts/kepler/download_data.py
IlayMalinyak: right. this was made for a simpler download but actually redundant. I will remove it.

scripts/kepler/download_data.py
```diff
@@ -0,0 +1,70 @@
...
+            shutil.rmtree(cache_dir)
+        
+        logging.info(f"Cleared cache for KIC {kic}")
+        return
+    except Exception as e:
+        logging.info(f"Error processing KIC {kic}: {e}")
+        return
+
+def download_data_parallel(args):
+    os.mkdir('./fits_data', exist_ok=True)
```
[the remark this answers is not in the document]
IlayMalinyak: True. will change.

scripts/kepler/download_data.py
IlayMalinyak: I ill update the function to get this as an argument

scripts/kepler/download_data.py
IlayMalinyak: True. will update

scripts/kepler/README.md
IlayMalinyak: this is the Kepler input catalog. It can be downloaded from https://archive.stsci.edu/pub/kepler/catalogs/
I will add it to the README.

scripts/kepler/build_parent_sample.py
```diff
@@ -0,0 +1,235 @@
...
+                times.append(time[good_data_mask])
+                sap_fluxes.append(sap_flux[good_data_mask])
+                pdcsap_fluxes.append(pdcsap_flux[good_data_mask])
+                sap_fluxes_errs.append(sap_flux_err[good_data_mask])
+                pdcsap_fluxes_errs.append(pdcsap_flux_err[good_data_mask])
+
+                
+
+    # Basic quality filtering
+    sap_fluxes = normalize_lightcurve(sap_fluxes)
```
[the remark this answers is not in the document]
IlayMalinyak: This normalization is per quarter - because Kepler measurements are divided into quarters (90 days measurement), and every quarter is independent, different quarters have different means. To stitch them together, I normalize each quarter to have a global mean (which is different among different samples). I thought this was basic enough to be done when making the dataset but if you think it's not necessary, I will remove it.

scripts/kepler/build_parent_sample.py
```diff
@@ -0,0 +1,235 @@
...
+                                np.isfinite(sap_flux_err) & \
+                                np.isfinite(pdcsap_flux) & \
+                                np.isfinite(pdcsap_flux_err)
+
+                times.append(time[good_data_mask])
+                sap_fluxes.append(sap_flux[good_data_mask])
+                pdcsap_fluxes.append(pdcsap_flux[good_data_mask])
+                sap_fluxes_errs.append(sap_flux_err[good_data_mask])
+                pdcsap_fluxes_errs.append(pdcsap_flux_err[good_data_mask])
+
```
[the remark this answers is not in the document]
IlayMalinyak: I will add. While reviewing this I noticed that not always all the fluxes appear in the fits file. In the case where the data is downloaded using the lightkurve package,  only the pdcsap flux is saved by default. I will handle this case also.

scripts/kepler/run.sh
IlayMalinyak: I might remove this entirely since this is system-specific

helenqu:
hey Ilay, thanks so much for your contribution and quick responses! as an update, we plan on incorporating Kepler data via your PR ASAP, so I'll be checking out your code to download and format the data and if everything works I'll be more than happy to merge it in.

IlayMalinyak:
He Helen,
It's me that needs to thank you for this beautiful project :)
I just updated the download_data method to be more robust. 
let me know if you have more comments.

IlayMalinyak:
I don't understand where in the Kepler dataset the tests failed. Regardless of that, currently, there is a manual step of creating a table with all the paths to the downloaded files. I can automate it as a post-download part.

helenqu:
the test seems to be failing for a different dataset so don't worry too much about that for now. 

I downloaded quarter 0 data from the bulk download site you listed, as well as a catalog from https://exoplanetarchive.ipac.caltech.edu/cgi-bin/TblView/nph-tblView?app=ExoTbls&config=q1_q17_dr25_koi. could you share the catalog creation script that makes the table you input into `build_parent_sample.py`? we're not very familiar with Kepler data so I'm not sure if the directory names in the download are the KIC/KID/kepID.

IlayMalinyak:
Hi,
KID, KIC, and kepid all mean the same id. In my script it expect to ba KID
(although I might generalize it to all names). The build_parent_sample
expect the same catalog you downloaded with an extra column named
'data_file_path' which should be a list of paths for each KID sample (list
because for multiple quarters you might have multiple paths. If you have
only one quarter create a list with one element). Does it make sense?

בתאריך שבת, 15 בפבר׳ 2025, 0:59, מאת Helen Qu ***@***.***>:

IlayMalinyak:
Hi,
is this still in progress?

mb010:
@helenqu do you have time to look at the changes or should I this week?

lhparker1:
I think the best thing to do would be to merge your newest changes into this branch, and then we can merge this one into main. Ping me when the newest changes are in and I can review it before we merge into main.

IlayMalinyak:
popping this up again

helenqu:
thanks for the reminder! I opened a new PR (https://github.com/MultimodalUniverse/MultimodalUniverse/pull/165) that incorporates all of your commits and my changes. If it looks good to you I'll go ahead and close this one.

PR #159, opened 2025-02-11
add SAGES DR1

al-jshen:
$u_s$ and $v_s$ photometry for ~48M stars from [Fan et al 2023](https://arxiv.org/abs/2306.15611)

"Its $v_s$-band filter is designed by the SAGES team covering the Ca II H&K lines, aiming to provide reliable stellar metallicity, while $u_s$ (similar to the $u$ band of Stromgren-Crawford system, covering the Balmer jump) could provide constraints on surface gravity for at least early type stars. The data can be used to derive the age and metallicity of stellar populations in the Milky Way, and nearby galaxies including M31/M33, as well as precise interstellar extinctions of individual stars."


al-jshen:
GH action fails because of unrelated issue (missing bz2), #161 should fix it.

scripts/sages/README.md
```diff
@@ -0,0 +1,11 @@
+# SAGES
+
+This is an implementation of an MMU compatible version of the Stellar Abundances and Galactic Evolution Survey (SAGES) DR1 dataset.
+
+## Data preparation
+
+Run the `download.py` script. This will download the catalog.
+
+Run the `process.py` script. This will process the catalog and create the necessary files in the MMU format.
```
lhparker1, line 9: as with galex, we should probably streamline these if possible

lhparker1 (commented):
LGTM other than combining the two scripts. Also can you add the details of the dataset to the front page README like with GALEX. Thanks!

scripts/sages/README.md
al-jshen: Added a convenience script to do both downloading and processing.

lhparker1 (approved):
LGTM!

PR #161, opened 2025-02-11
fix missing bz2 for fitsio install in GH action

al-jshen:
This is causing the action to fail in the dependency install stage before any real tests are run.


mb010 (approved):
LGTM

PR #163, opened 2025-02-24
bug fix for 2mass

al-jshen:
add missing dec key (is called `decl`)

### DIFF
--- a/scripts/twomass/test_load.py
+++ b/scripts/twomass/test_load.py
@@ -9,6 +9,10 @@ ds_hdf5 = datasets.load_dataset(
 
 batch_hdf5 = next(iter(ds_hdf5))
 
+print(batch_hdf5["object_id"])
+print(batch_hdf5["ra"])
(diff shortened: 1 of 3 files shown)


scripts/twomass/test_load.py
```diff
@@ -9,6 +9,10 @@
 
 batch_hdf5 = next(iter(ds_hdf5))
 
+print(batch_hdf5["object_id"])
+print(batch_hdf5["ra"])
+print(batch_hdf5["dec"])
```
lhparker1, line 14: do we want these prints still here?

al-jshen: This is just a test loading script so it only ever is run during the github action. So I think it is fine to have (as a check that those 3 keys exist).

lhparker1: Ok :)

lhparker1 (approved):

PR #174, opened 2025-07-15
LAMOST stellar spectra

IlayMalinyak:
Hi,
I created a dataset for LAMOST spectra. LAMOST consists of millions of stellar spectra and I find it very improtant for stellar multimudality models. 
Note that I didn't implement any quality cuts (as opposed to the APOGEE dataset, for example). This allows users to use their 
quality cuts.

Ilay


al-jshen:
Thanks @IlayMalinyak for the submission! I have a couple comments/suggestions:

1) This seems at the moment to be written specifically for the LRS stellar parameter catalog, but LAMOST has a few different catalogs available. While most of the keys are the same between the catalogs, there are a few that are specific to certain catalogs. For example `rv` is only in the stellar parameter catalog, but not in the general catalog, and `mg_fe` is in the MRS stellar parameter catalog, but not in the LRS stellar parameter catalog. So when I tested your build script it failed for the general catalog. I suggest having different catalogs of the dataset supported through different `BUILDER_CONFIGS`, where each one has both the common keys and also the catalog-specific keys listed. Then a user can specify which catalog they want to use and this will work with all the catalogs.

2)  Can we add some sort of automated download for the catalog file? They aren't blocked behind any authentication and I am able to just directly `wget` them. I suggest adding two flags to the `download_data.py` script, one for the DR version and one to specify which catalog, and it will automatically download the catalog file if it does not exist already. This will simplify and streamline the download process.

IlayMalinyak:
@al-jshen thanks for the valuable feedback!
I had only the LRS stellar catalog in my mind because this is what I'm using, but we can definitely use all the catalogs with some BUILDER_CONFIG, as you suggested. Automated download is also a good point and should be possible. I'm a little bit busy these days, but I will get to it and fix those issues.
Ilay

al-jshen:
No worries! I've added the automated catalog download, will get to the BUILDER_CONFIG thing later

al-jshen:
Turns out MRS is structured quite a bit differently, I will just do LRS for now.

al-jshen:
Alright, I've made the proposed changes. Let me know what you think @IlayMalinyak.

IlayMalinyak:
Hi @al-jshen
Thanks for the update! That looks good. I just removed an unnecessary function. From my side, we can merge this.

al-jshen (approved):
looks good to me, thanks @IlayMalinyak for the nice work!

